PR #5, opened 2023-07-27
Fix streaming for llama 7b.

squidarth:
Streaming was broken for the Llama 7b streaming truss, this fixes it.


philipkiely-baseten (approved):
Cool, should the same change be applied across the other 5 variants?

I can update model library when this lands.

squidarth:
> Cool, should the same change be applied across the other 5 variants?

> I can update model library when this lands.

yes! i can tackle that tomorrow

PR #6, opened 2023-07-29
model library style guide

philipkiely-baseten:
Getting us started with a style guide for model library models and example Trusses that we can expand over time to make sure models are consistent and demonstrate best practices.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,73 @@
+# Contributing
+
+Please open a PR to add a new model.
+
+## Style guide
+
+Model library models should follow this style guide.
+
+General style tips:
+
+* Truss folder name should use hyphens, not underscores (e.g. `falcon-7b` not `falcon_7b`)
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,73 @@
...
+- sentencepiece==0.1.99
+- torch==2.0.1
+- transformers==4.30.2
+```
+
+### Secrets
+
+Model library models can access secrets.
+
+If the model requires HuggingFace (e.g. Llama 2), always call the secret `hf_access_token`
```
bolasim, line 55: cc: @varunshenoy

bolasim (approved):
This is awesome. Thank you so much for writing this. 

Do you have opinions on how we should organize the top level of this repo? Maybe by use case or model family group or something?

Also, Would love to write a pre-commit linter for trusses. I might take a pass at it to make sure this style guide is enforced.

CONTRIBUTING.md
philipkiely-baseten: Will think RE top level organization.

PR #7, opened 2023-08-02
Batching with Stable Diffusion XL

varunshenoy:
SCR-20230801-pihy


stable-diffusion-xl-1.0/config.yaml
```diff
@@ -12,16 +12,24 @@ model_name: Stable Diffusion XL
...
+  - vae/diffusion_pytorch_model.safetensors
 requirements:
 - transformers
 - accelerate
 - safetensors
 - diffusers
 - invisible-watermark>=0.2.0
 resources:
-  accelerator: A10G
+  accelerator: A100
```
squidarth, line 30: @varunshenoy does this need A100? I think ideally models in truss-examples don't require it  (since not all users will have access)

amiruci: hopfully very soon we can enable it for everyone btw.

PR #9, opened 2023-08-11
Add SDXL example with streaming

bolasim:


sdxl-1.0-stream/model/model.py
```diff
@@ -0,0 +1,84 @@
...
+            b64_img = self.convert_to_b64(image)
+            q.put(b64_img + "\n")
+        
+        def generation_task():
+            image = self.pipe(prompt=prompt, output_type="latent" if use_refiner else "pil", callback=latents_callback, callback_steps=15).images[0]
+            
+            if use_refiner:
+                image = self.refiner(prompt=prompt, image=image[None, :]).images[0]
+            final_image = self.convert_to_b64(image)
+            q.put(final_image + "\n")
```
squidarth, line 68: worth leaving a comment explaining why we are doing this

sdxl-1.0-stream/model/model.py
```diff
@@ -0,0 +1,84 @@
...
+            q.put(job_done)
+        
+        thread = Thread(
+            target=generation_task,
+        )
+        thread.start()
+        def inner_cosumer():
```
squidarth, line 75: spelling

sdxl-1.0-stream/model/model.py
```diff
@@ -0,0 +1,84 @@
...
+            while True:
+                next_item = q.get(True, timeout=300) # Blocks until an input is available
+                if next_item is job_done:
+                    break
+                logging.warn(len(next_item))
+                yield next_item
```
squidarth, line 81: outside of this `while` loop, add a thread.join(). This will prevent the thread from getting garbage collected (after `predict` returns, there are no more references to it.

squidarth (approved):

sdxl-1.0-stream/model/model.py
```diff
@@ -0,0 +1,95 @@
...
+        def latents_callback(i, t, latents):
+            if i < 25:
+                return
+            
+            # make sure the VAE is in float32 mode, as it overflows in float16
+            if vae.dtype == torch.float16 and vae.config.force_upcast:
+                self.pipe.upcast_vae()
+                latents = latents.to(next(iter(vae.post_quant_conv.parameters())).dtype)
+
+            image = vae.decode(latents / vae.config.scaling_factor, return_dict=False)[0]
```
squidarth, line 64: You need to ensure that `q.put(job_done)` happens in all cases. Consider using `try/finally` to ensure that this happens in the case of exceptions.

squidarth (approved):
if we want to add this still, could we move to `stable-diffusion` folde?

PR #17, opened 2023-09-16
audiogen readme

philipkiely-baseten:
How do we want users to do this without Baseten client? Should they just be using the requests library? Or is this what the `-f` flag is for on `truss predict`? Let's make musicgen melody one of our examples because it uses files for both input and output.


squidarth (approved):
Alternative would be do something like this:
```
import requests

url = "https://app.baseten.co/models/YOUR_MODEL_ID/predict"
headers = {
    'Authorization': 'Api-Key YOUR_API_KEY'
}

# Define the data payload
data = {"prompts": ["happy rock" "energetic EDM", "sad jazz"], "melody": encoded_str, "duration": 8}

# Make the POST request
response = requests.post(url, headers=headers, data=data)

data = response.json()["data"]
for idx, clip in enumerate(data):
    ...
```

PR #18, opened 2023-09-22
add max new tokens

philipkiely-baseten:
Add `max_new_tokens` to example in response to user feedback. I had to use this parameter for Chainlit and LangChain to work properly as I was using longer prompts.


amiruci:
i'm curious, did it work well when passed the max of 4096? if so, ship it.

philipkiely-baseten:
Yep, it worked well! It does not generate 4096 tokens, only enough to answer the prompt, which is the expected and desired behavior.

PR #20, opened 2023-09-26
Adding truss for code llama 7b

htrivedi99:


code-llama-7b-instruct/model/model.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def __init__(self, data_dir: str, config: Dict, **kwargs):
+        self._data_dir = data_dir
+        self._config = config
+        self.model = None
+        self.tokenizer = None
+
+    def preprocess(self, request: dict):
+        generate_args = {
+            "max_new_tokens": 128,
```
squidarth, line 22: nit: could we have this be:
```
"max_new_tokens": request.get("max_tokens", 128)
```

instead? (and ditto for the other parameters)

code-llama-7b-instruct/model/model.py
```diff
@@ -0,0 +1,69 @@
...
+                formatted_prompt = f"<s><<SYS>>\n{DEFAULT_SYSTEM_PROMPT}\n<</SYS>>\n\n{prompt}"
+                input_ids = self.tokenizer(
+                    formatted_prompt, return_tensors='pt').input_ids.cuda()
+                output = self.model.generate(
+                    inputs=input_ids,
+                    **request["generate_args"]
+                )
+
+                return self.tokenizer.decode(output[0])
+            except Exception as exc:
```
squidarth, line 68: I don't think we need this try/except -- exceptions will be handled by Truss and turned into the correct errors

squidarth (commented):
This looks great! Left a couple nits.

One q: just to double check, code llama doesn't require some kind of special license, does it?

code-llama-7b-instruct/model/model.py
htrivedi99: Good suggestion. I updated the code.

code-llama-7b-instruct/model/model.py
htrivedi99: Ok, try except block removed

htrivedi99:
> This looks great! Left a couple nits.
> 
> One q: just to double check, code llama doesn't require some kind of special license, does it?



> This looks great! Left a couple nits.
> 
> One q: just to double check, code llama doesn't require some kind of special license, does it?

When you say special license do you mean permission to access the model?

squidarth:
> When you say special license do you mean permission to access the model?

Yeah -- exactly

code-llama-7b-instruct/model/model.py
```diff
@@ -0,0 +1,59 @@
...
+        )
+        self.tokenizer = AutoTokenizer.from_pretrained(
+            self.MODEL_NAME
+        )
+
+    def predict(self, request: Dict) -> Dict:
+        with torch.no_grad():
+            prompt = request.pop("prompt")
+            # Code-llama needs the prompt to be formatted in this manner
+            formatted_prompt = f"<s><<SYS>>\n{DEFAULT_SYSTEM_PROMPT}\n<</SYS>>\n\n{prompt}"
```
varunshenoy, line 50: Can we also template out the default system prompt? Would be good to give users a way to pass in their own system prompt.

varunshenoy (commented):
Awesome, just a small nit.

htrivedi99:
> > When you say special license do you mean permission to access the model?
> 
> Yeah -- exactly

No, code llama does not require any special license

squidarth:
Great! thanks again for the contributioN!

PR #21, opened 2023-09-27
Add vLLM example

aspctu:
This PR adds a quick example on how to use vLLM inside of Truss

### DIFF
--- a/wizard-vllm/config.yaml
+++ b/wizard-vllm/config.yaml
@@ -4,9 +4,9 @@ live_reload: false
 model_name: wizard_vllm
 python_version: py311
 requirements:
-- vllm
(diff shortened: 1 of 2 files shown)


wizard-vllm/config.yaml
```diff
@@ -0,0 +1,19 @@
+description: null
+environment_variables: {}
+live_reload: false
+model_name: wizard_vllm
+python_version: py311
+requirements:
+- vllm
+- ray
+- pyarrow
+- pandas>=1.3
```
varunshenoy, line 10: small nit: pin the versions?

varunshenoy (approved):
lgtm!

PR #23, opened 2023-09-27
Added LoRA Swapping Truss

varunshenoy:
Adds an example Truss that shows how LoRAs can be hot swapped within `predict`.

### DIFF
--- a/sdxl-lora-swapping/model/model.py
+++ b/sdxl-lora-swapping/model/model.py
@@ -1,0 +1,96 @@
+from typing import Any
+
+from diffusers import DiffusionPipeline, AutoencoderKL
+import torch
+import base64
+from io import BytesIO
+from PIL import Image
+import time
+import functools
+
+
+# Good notebook to learn how to use diffusers + LoRA:
+# https://colab.research.google.com/gist/sayakpaul/109b7e792c64514fb3c057ecff4145ff/scratchpad.ipynb#scrollTo=GreOMxAcvlm8
+
+class Model:
+    def __init__(self, **kwargs) -> None:
+        self.pipe = None
+
+    def load(self):
+        # Load model here and assign to self._model.
+        vae = AutoencoderKL.from_pretrained(
+            "madebyollin/sdxl-vae-fp16-fix", torch_dtype=torch.float16
+        )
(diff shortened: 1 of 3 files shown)


sdxl-lora-swapping/model/model.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+# Good notebook to learn how to use diffusers + LoRA:
+# https://colab.research.google.com/gist/sayakpaul/109b7e792c64514fb3c057ecff4145ff/scratchpad.ipynb#scrollTo=GreOMxAcvlm8
+
+class Model:
+    def __init__(self, **kwargs) -> None:
+        self.pipe = None
+
+    def load(self):
```
joostinyi, line 19: In cases where the kind and number of adapters + their size is small enough to fit in CPU RAM, is it still more performant to load these from remote to gpu mem rather than loading all in `load()` and on/offloading upon requests?

sdxl-lora-swapping/model/model.py
```diff
@@ -0,0 +1,96 @@
...
+        high_noise_frac = model_input.pop("high_noise_frac", 0.8)
+        num_inference_steps = model_input.pop("num_inference_steps", 30)
+
+        lora = model_input.pop("lora", None)
+        print(f"Loading LoRA weights: {lora}")
+
+        # example schema: 
+        # {"repo_id": "nerijs/pixel-art-xl", "weights": "pixel-art-xl.safetensors"}
+
+        if lora is not None:
```
joostinyi, line 69: Does this always download from remote if `lora` kwarg is in the input? What if the correct adapter is already loaded in gpu memory from previous request?

sdxl-lora-swapping/model/model.py
varunshenoy: if you have adapters, and you know them ahead of time, you should just use `hf_cache` or pop them on at a time in the load if you're ok with downloading the adapters.

sdxl-lora-swapping/model/model.py
varunshenoy: It won't redownload if it was used last. The default behavior here is to use the last loaded LoRA since it modifies the `pipe` itself.

PR #29, opened 2023-10-03
Adding control net qr code model

htrivedi99:


control-net-qrcode/README.md
```diff
@@ -0,0 +1,60 @@
...
+truss push
+```
+
+Paste your Baseten API key if prompted.
+
+For more information, see [Truss documentation](https://truss.baseten.co).
+
+### Hardware notes
+
+This two models combined only take up about 5 GB of VRAM so a T4 is enough for this truss.
```
varunshenoy, line 33: How fast is it on a T4? I'd probably recommend bumping the default to an A10G.

control-net-qrcode/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Hardware notes
+
+This two models combined only take up about 5 GB of VRAM so a T4 is enough for this truss. 
+
+### API route: `predict`
+
+The predict route is the primary method for generating images based on a given prompt. It takes several parameters:
+
+- __prompt__ (required): The input text required for image generation.
```
varunshenoy, line 39: is there a parameter for the QR Code or (alternatively) a mask image?

control-net-qrcode/config.yaml
```diff
@@ -0,0 +1,21 @@
+environment_variables: {}
+external_package_dirs: []
+model_metadata: {}
+model_name: control-net-qrcode
+python_version: py310
+requirements:
```
varunshenoy, line 6: thank you for pinning version numbers!

control-net-qrcode/model/model.py
```diff
@@ -0,0 +1,103 @@
...
+        self.model = StableDiffusionControlNetPipeline.from_pretrained(
+            "runwayml/stable-diffusion-v1-5",
+            controlnet=controlnet,
+            safety_checker=None,
+            torch_dtype=torch.float16,
+        ).to("cuda")
+        self.model.enable_xformers_memory_efficient_attention()
+
+    def predict(self, request: Dict) -> Dict:
+        qr_code_content = request.get("qr_code_content")
```
varunshenoy, line 67: the content can be either:
- a url to turn into a qr code
- OR more generally, some kind of mask so users can make images like this: https://twitter.com/MrUgleh/status/1702041188482658758?s=20

varunshenoy (commented):
Looks good! Just a few questions + changes. The .gif is a fantastic touch.

control-net-qrcode/README.md
htrivedi99: It seems to be fast enough on the T4 so I don't think the A10G is needed, especially since most of the VRAM will be unused.

control-net-qrcode/README.md
htrivedi99: There is a way to provide a custom image as a mask, but then it won't generate a QR code. Do you think it would be better to have 2 different truss models, one for the QR code generator and another for custom images?

varunshenoy: I think it can be the same truss. You can branch based on a bool/key in the request.

htrivedi99: Ok, I've added a new key to read the mask if the qr_code_content is not provided

squidarth:
@htrivedi99 just a heads up, we've restructured the truss-examples repo a bit, such that the top level are models that are connected to our docs (the numbered examples). Everything else moves to the "model_library" (we're going to be adding some more structure to that folder too, going forward

htrivedi99:
> @htrivedi99 just a heads up, we've restructured the truss-examples repo a bit, such that the top level are models that are connected to our docs (the numbered examples). Everything else moves to the "model_library" (we're going to be adding some more structure to that folder too, going forward

Sounds good!

PR #34, opened 2023-10-10
Adding ExLlama V2 Examples

varunshenoy:
This PR adds two Trusses that showcase how to use ExLlama V2. One streams while the other is a standard generation model.


squidarth (approved):
overall lgtm, just a couple

PR #36, opened 2023-10-11
Adding truss for real-esrgan

htrivedi99:


squidarth (commented):
hm, should we pin the ersgan version to a particular SHA? (Just to keep this consistent)?  Code overall looks good though

htrivedi99:
> hm, should we pin the ersgan version to a particular SHA? (Just to keep this consistent)? Code overall looks good though

Cool, I pinned the repo to a specific commit

htrivedi99:
@squidarth Sure, I can rearrange it. 
Looking at the new repo, the folders labeled 1-10 look great, but the rest of the models seem a bit scattered. Do you think there is any value in binning these models by category like: "LLM", "Audio Generation", "Super Resolution", "Image Generation", etc?

PR #38, opened 2023-10-12
Adding speaker diarization to whisperx

htrivedi99:


model_library/whisperx-truss/config.yaml
```diff
@@ -21,7 +21,8 @@ resources:
   memory: 14Gi
   use_gpu: true
   accelerator: T4
-secrets: {}
+secrets:
+  hf_access_token: "YOUR-HUGGING-FACE-TOKEN"
```
squidarth, line 25: let's just make this null `hf_access_token: null` for now -- it'll give us a better error message

squidarth (approved):
awesome, thanks for doing this! One quick thing: in the config, under `model_metadata`, you can add the `example_model_input` field, that's often helpful for understanding how to invoke a model.

```
model_metadata:
  example_model_input: {"prompt": "What is the meaning of life?"}
```

model_library/whisperx-truss/config.yaml
htrivedi99: Updated config to reflect these suggestions

PR #46, opened 2023-10-23
Fix mistral-7B models in examples

ttim:
I've discovered that mistral-7B model doesn't work currently due to typo and using model class instead of tokenizer in `model.py`.

I've fixed this, as well as enabled two mistral models on ci tests to catch this error next time.

### DIFF
--- a/ci.yaml
+++ b/ci.yaml
@@ -7,3 +7,5 @@ tests:
   - 06-high-performance-cached-weights
   - 07-high-performance-tgi
   - 10-using-system-packages
+  - mistral/mistral-7b
+  - mistral/mistral-7b-instruct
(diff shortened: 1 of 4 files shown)


squidarth (approved):
thanks for fixing!

ci.yaml
```diff
@@ -6,4 +6,6 @@ tests:
   - 05-speech-to-text
   - 06-high-performance-cached-weights
   - 07-high-performance-tgi
-  - 10-using-system-packages
\ No newline at end of file
+  - 10-using-system-packages
```
varunshenoy, line 9: Think we should leave out specific models from this?

cc @squidarth

varunshenoy (approved):
Looks good! Just pinging Sid on the CI edit.

ci.yaml
varunshenoy: nvm, looks good

ttim:
Failed without my change: https://github.com/basetenlabs/truss-examples/actions/runs/6617979511/job/17975438184
Succeed with: https://github.com/basetenlabs/truss-examples/actions/runs/6617984796

PR #52, opened 2023-10-26
Add Jina Embedding model

philipkiely-baseten:
Add a Truss for Jina embedding model, intending to add to model library today + publish a blog post on this


jina-embeddings/jina-embeddings-v2-base-en/model/model.py
```diff
@@ -0,0 +1,16 @@
+from transformers import AutoModel
+
+class Model:
+    def __init__(self, **kwargs):
+        self._model = None
+
+    def load(self):
+        self._model = AutoModel.from_pretrained(
```
bolasim, line 8: bonus, you can use `hf_cache` in the config with these settings so the cold start is fire out of the box.

squidarth (approved):
awesome! could you also please add this to the [ci.yaml](https://github.com/basetenlabs/truss-examples/blob/main/ci.yaml)? lmk if you need help getting the lint to pass

jina-embeddings/jina-embeddings-v2-base-en/model/model.py
philipkiely-baseten: Added!

jina-embeddings/jina-embeddings-v2-base-en/config.yaml
```diff
@@ -0,0 +1,27 @@
+environment_variables: {}
+external_package_dirs: []
+hf_cache:
+  - repo_id: jinaai/jina-embeddings-v2-base-en
```
bolasim, line 4: you probably wanna pin the revision here just like you're doing in the load.

philipkiely-baseten: Ah, makes sense, do you know how to do that? Would it just be `revision`: `abcd1234`?

bolasim: https://truss.baseten.co/reference/config#hf-cache-list-item-revision

philipkiely-baseten:
@bolasim I was having trouble getting the `hf_cache` working with the revision, and I want to get this live, so I took `hf_cache` out entirely when I updated to the latest version.

The model is a quarter of a gig, so the download actually only takes about a second (log line below):

```
Downloading model.safetensors: 275MB [00:01, 264MB/s]
...
Completed model.load() execution in 8419 ms
```

The whole model.load() is 8.5 seconds which is not great but not terrible. And the model is running faster now with the new revision, I'm getting 40-45 seconds for my shakespeare dataset.

bolasim:
Whatever you think is best!

PR #55, opened 2023-10-31
Adding TRT-LLM + Triton truss

aspctu:


trtllm-truss/config.yaml
```diff
@@ -0,0 +1,16 @@
...
+  tensor_parallelism: 1
+model_name: trtllm-triton
+python_version: py311
+requirements: []
+resources:
+  accelerator: A100
+  use_gpu: true
+secrets: {}
+system_packages: []
+predict_concurrency: 256
```
bolasim, line 16:
```suggestion
runtime:
    predict_concurrency: 256
```

bolasim: I am surprised this didn't through a validation error when you tried to upload

trtllm-truss/model/model.py
```diff
@@ -0,0 +1,74 @@
...
+        self._tensor_parallel_count = self._config["model_metadata"].get("tensor_parallelism", 1)
+        self.triton_client = TritonClient(self._data_dir, TRITON_MODEL_REPOSITORY_PATH, self._tensor_parallel_count)
+        self.triton_client.load_server_and_model(
+            env={
+                "HUGGING_FACE_HUB_TOKEN": self._secrets["hf_access_token"]
+                "triton_tokenizer_repository": self._config["model_metadata"]["tokenizer_repository"],
+            }
+        )
+
+    def predict(self, model_input):
```
pankajroark, line 29: We should try to use async predict. Sync predict runs on a thread pool which has limited number of threads and can limit concurrency. Plus creating a new thread per request is not ideal. cc @squidarth who may know of examples of where we use async predict.

pankajroark: Ah, I missed the yield before. So this predict function is a generator, right?

squidarth: it's not obvious to me that there would be a big perf increase by switching to having this by async (it's true that doing things this way will produce another thread). It's a medium lift at least to switch it, since we'd have to change the TritonClient implementation to also be async

trtllm-truss/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+```
+parameters: {
+  key: "kv_cache_free_gpu_mem_fraction"
+  value: {
+    string_value: "0.1"
+  }
+}
+```
+TODO(Abu): __fill__
```
joostinyi, line 64: for your [reference](https://github.com/triton-inference-server/tensorrtllm_backend/blob/47b609b670d6bb33a5ff113d98ad8a44d961c5c6/inflight_batcher_llm/src/libtensorrtllm.cc#L1113): looks like using both `max_tokens_in_paged_kv_cache` and `kv_cache_free_gpu_mem_fraction` are redundant. I didn't find docs on `kv_cache_free_gpu_mem_fraction` but sounds like they preallocate 85% of the free gpu mem for the kv cache by default, if `max_tokens_in_paged_kv_cache` is not specified

joostinyi (approved):

PR #59, opened 2023-11-08
Adding whisper v3 truss

htrivedi99:


squidarth (approved):
Awesome, looks great, thanks for doing this @htrivedi99!

We added linting recently, you can install the pre-commit hooks with ` pre-commit install`

PR #65, opened 2023-11-10
WIP

htrivedi99:


xtts-v2-truss/config.yaml
```diff
@@ -0,0 +1,17 @@
+environment_variables: {}
```
bolasim, line 1: you can drop the keys you're note using (like `environment_variables`  and `external_package_dirs` and `python_version` since you're using a base image)

xtts-v2-truss/model/model.py
```diff
@@ -0,0 +1,75 @@
...
+
+from TTS.api import TTS
+
+
+class Model:
+    def __init__(self, **kwargs):
+        self.model = None
+
+    def load(self):
+        os.environ["COQUI_TOS_AGREED"] = "1"
```
bolasim, line 13: you can actually set this in the `config.yaml` instead. that might be clearer

bolasim (approved):

xtts-v2-truss/config.yaml
```diff
@@ -0,0 +1,15 @@
+environment_variables:
+  COQUI_TOS_AGREED: "1"
+model_metadata:
+  example_model_input: {"text": "I love robots. Robots are cool!", "speaker_voice": <BASE64-STRING>, "language": "en"}
+  tags:
+  - text-to-speech
+base_image:
+  image: bolabaseten/xtts-base:v0.0.1
```
bolasim, line 8: @htrivedi99 do we still need this or can we just add XTTS as a depedency?

PR #69, opened 2023-11-16
Add `ignore_eos` support to deep-speed and trt-llm backends

ttim:
In order to make benchmarks more apples-to-apples it would be useful to have an ability to keep generating output even if `eos` token was emitted. I've added support for that to deep speed and trt-llm backends.

### DIFF
--- a/llama/llama-2-7b-trt-llm/model/model.py
+++ b/llama/llama-2-7b-trt-llm/model/model.py
@@ -1,6 +1,7 @@
 from itertools import count
 from pathlib import Path
 from threading import Thread
+from transformers import AutoTokenizer
 
 import numpy as np
 from client import TritonClient, UserData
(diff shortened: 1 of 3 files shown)


llama/llama-2-7b-trt-llm/model/model.py
```diff
@@ -61,7 +68,11 @@ def predict(self, model_input):
         bad_words_list = model_input.get("bad_words_list", [""])
         stop_words_list = model_input.get("stop_words_list", [""])
         repetition_penalty = model_input.get("repetition_penalty", 1.0)
-        end_id = model_input.get("end_id", 2)  # EOS token for Llama2
+        if not model_input.get("ignore_eos", False):
+            end_id = model_input.get("end_id", self.eos_token_id)
+        else:
+            # do nothing, trt-llm by default doesn't stop on `eos`
+            pass
```
joostinyi, line 75: In this case isn't `end_id` undefined but we attempt to prepare and pass this tensor to the server?

ttim: Yeah, it was something weird written before, thanks for the catch!

joostinyi (approved):

joostinyi:
We will probably need to revisit this in the future – the default functionality depends on unintended behavior on upstream TRT-LLM so we should be able to explicitly specify full output length generation

PR #71, opened 2023-11-17
Add support of `ignore_eos` to mistral vllm truss 

ttim:
- Add support of `ignore_eos` 
- Add support of `stream`
- Add support of customizable prompts format & repo_id


joostinyi (approved):
Changes look good to me – only thing for @aspctu to chime in on, do you think we should do better to stay up to date with vllm upstream by pinning to the most recent pip installable `vllm`?

ttim:
@joostinyi changed to `requirements` and also simplified bunch of simplifications.

pankajroark (approved):

PR #72, opened 2023-11-17
Use vllm api in deepspeed and trt-llm

ttim:
- Mimic vllm api in deepspeed truss
- Mimic vllm api in llama 2 7b trt-llm truss
- Support streaming in trt-llm truss


pankajroark (approved):

amiruci:
are there any READMEs that should change as part of this?

ttim:
@amiruci good catch, updated READMEs

PR #73, opened 2023-11-17
Adding Mistral TRT-LLM truss

aspctu:


ttim:
How different this truss is from Llama one? Can we instead do changed to llama one and guard it with flag?

pankajroark (approved):

aspctu:
@ttim It's not too different except for the fact that it uses a completely new base image (with the latest `main` for TRT-LLM).

PR #74, opened 2023-11-20
Add a mistral chat model.

squidarth:
Main difference between this and the instruct model is that the inputs to this are a list of dictionaries of the format:

```
[
  {"role": "user", "content": "What is the mistral wind?",
  {"role": "assistant", "content": "blah blah blah"...
]

```


mistral/mistral-7b-chat/config.yaml
```diff
@@ -0,0 +1,25 @@
+environment_variables: {}
```
ttim, line 1: Let's remove all the options equal to default values?

ttim (approved):
Do we want to add `open-ai-api` into the name?

PR #78, opened 2023-11-21
Add model server build config  

joostinyi:
- Adds model server build config to prevent development deployment of TRT-LLM servers, while we work on adding patching support to these examples


bolasim (approved):

ttim:
This change breaks truss, I assume until https://github.com/basetenlabs/truss/pull/748 is released. Can we revert it so far.

PR #82, opened 2023-11-21
Fix zephyr & mistral trusses

squidarth:
Fix zephyr & mistral trusses.

These should've been keys
Zephyr_7B_Alpha___Model___Baseten


bolasim (approved):
ma bad

ttim (approved):

PR #124, opened 2023-12-12
add on the fly engine build documentation

joostinyi:


mistral/mistral-7b-trt-llm-build-engine/README.md
```diff
@@ -6,6 +6,7 @@ This is a [Truss](https://truss.baseten.co/) for Mistral 7B. This README will wa
 
 This truss differs from `mistral-7b-trt-llm` by supporting building trt-llm engines on the fly on the model load.
 This significantly increases model start time and therefore it's **not intended for production use**.
+To make use of this feature, specify both the build command in `model_metadata.engine_build` and huggingface repository with your weights and tokenizer in`model_metadata.tokenizer_repository`
```
ttim, line 9: Worth nothing that you don't need to specify either output or model directories, they will be supplied automatically

joostinyi: I'm hesitant to add too many details about the build command args, especially for those looking to use the example without making any edits. Adding a note to check the file with implementation for those that need to see the details.

ttim (approved):

PR #128, opened 2023-12-13
Style tts 2

htrivedi99:


bolasim (commented):
please remove the StyleTTS2 folder from this repo. You can clone it at runtime if there is no other way, but ideally, uou can just use the `git+https://` in the requirements of the config.yaml

PR #134, opened 2023-12-18
mixtral 8x7b - trtllm

pankajroark:
One truss for the trtllm version of the original fp16 model. Another one with weight-only quantization. Weight-only quantized version can run on a single A100, which is really convenient.

Main change is in `templates/generate.yam` and the two README files in the respective truss folders.


ttim (approved):
Worth documenting max input/output lengths for these models. Other than that LGTM. Exciting!

PR #137, opened 2023-12-22
Avoid downloading engine more than once

ttim:
In cases when we run TRT-LLM trusses with more than one worker we download engine files each time worker runs. I've added safeguard on this so we download files only once.


pankajroark (approved):
awesome. The double download has been so cumbersome during testing, happy to see this fixed.
I assume we've done a test run with the new Truss?

ttim:
@pankajroark yeah, I've tested truss both with and without building engine. Seems working correctly.

PR #138, opened 2024-01-02
adding truss for nous capybara 34b

htrivedi99:


philipkiely-baseten (commented):
Does the user want streaming output for this model? I think by default we should make LLMs have a streaming capability unless the user really doesn't care about that.

Also, the example prompt included is not an instruct-style prompt, while the Hugging Face page shows a more conversation-style prompt template: https://huggingface.co/NousResearch/Nous-Capybara-34B#prompt-format

Other than that this looks good, let me know if either you make the changes or determine that they're not needed and I'll approve and merge.

htrivedi99:
Ok, I'll add the streaming functionality and make the prompt more instruct-style

philipkiely-baseten (approved):
Thanks for making these changes!

PR #143, opened 2024-01-05
Do not use pytorch in trt llm trusses

ttim:
I used logic from https://github.com/triton-inference-server/tensorrtllm_backend/blob/6e6e34e0944c10a82ad0bd526087d72ce16b8d56/all_models/inflight_batcher_llm/preprocessing/1/model.py to pad inputs with `numpy` instead of using `pytorch` which is heavy dependency we don't really need.


joostinyi (approved):
looks good!

pankajroark (approved):

PR #146, opened 2024-01-09
Remove clone from each model startup for ComfyUI

bolasim:
First pass at cleaning up this example. The should probably make it easy to add the models to the cache but will just take it step by step and hopefully invite other people to contribute.

@htrivedi99 would you mind taking a look please? I tested the dev deployment.


htrivedi99:
looks good to me :) 
I ran the truss as well just to make sure everything still works

htrivedi99 (approved):
lgtm

PR #148, opened 2024-01-10
two text embedding models

philipkiely-baseten:


all-minilm-l6-v2/config.yaml
```diff
@@ -0,0 +1,16 @@
...
+external_package_dirs: []
+model_metadata:
+  example_model_input: {"text": ["I want to eat pasta", "I want to eat pizza"]}
+model_framework: custom
+model_name: all-miniLM-L6-v2
+python_version: py39
+requirements:
+- sentence-transformers==2.2.2
+resources:
+  accelerator: null
```
squidarth, line 11: if it's null we can just exclude it from the config

all-mpnet-base-v2/config.yaml
```diff
@@ -0,0 +1,16 @@
...
+external_package_dirs: []
+model_metadata:
+  example_model_input: {"text": ["I want to eat pasta", "I want to eat pizza"]}
+model_framework: custom
+model_name: all-mpnet-base-v2
+python_version: py39
+requirements:
+- sentence-transformers==2.2.2
+resources:
+  accelerator: null
```
squidarth, line 11: ditto

squidarth (approved):

PR #149, opened 2024-01-10
Remove all deploy buttons

relph:
This PR removes all the `Deploy` buttons from model READMEs. These will likely return in some form after we transition to the new model library but until then lets remove them so the links aren't broken for each of these models.


squidarth (approved):
thanks for doing this

PR #151, opened 2024-01-12
Example Background Task Processor for long running predictions

bolasim:


11-misc/background-queue/config.yaml
```diff
@@ -0,0 +1,16 @@
+environment_variables: {}
+external_package_dirs: []
+model_metadata: {}
+model_name: Long Running Predictions
+python_version: py311
+requirements:
+- aiohttp
```
squidarth, line 7: does not look like we're using this

11-misc/background-queue/model/model.py
```diff
@@ -0,0 +1,66 @@
...
+            print(f"POST: {response.status_code}, text: {response.text}")
+
+import time
+
+class Sleeper:
+    def __call__(self, data):
+        for i in range(6):
+            print(f"task: {data}, processing {i+1}/6")
+            time.sleep(60)  # sleep for 1 minute
+        print("Done with data.")
```
squidarth, line 38: let's have this return some data so there's a payload to print

11-misc/callback_server.py
```diff
@@ -0,0 +1,42 @@
+from http.server import BaseHTTPRequestHandler, HTTPServer
+import logging
+
+class S(BaseHTTPRequestHandler):
```
squidarth, line 4: i know it requires installing something, but I think this would be more readable / easy to follow as a flask server. Maybe we can have setup instructions at the top or in a readme here

squidarth (approved):

PR #157, opened 2024-01-16
Removing A100 requirement from controlnet trusses

htrivedi99:


squidarth (approved):
looks good! (Assuming this works)

PR #158, opened 2024-01-17
stable diffusion xl 1.0 -- tensorrt version

pankajroark:
Based on https://huggingface.co/stabilityai/stable-diffusion-xl-1.0-tensorrt/blob/main/README.md#usage-example with certain limitations:
1. guidance_scale has to be > 1.0, i.e. only the classifier-free guidance mode is supported right now.
2. One prompt at a time, no batching.
3. Refiner is always used, configurable using the denoising_end parameter, which decides the fraction of steps that are run on refiner vs base model.
4. use of cuda graphs is disabled as using cuda graphs wasn't working well.

Tested it locally, will test it on Baseten soon.


joostinyi (commented):
One thing that wasn't clear to me was how the warmup runs are conducted that will affect perf. Is this unnecessary when not enabling cuda graphs?

bolasim (approved):

PR #160, opened 2024-01-17
downgrading resource requirements for a few popular models

htrivedi99:


bolasim (approved):
can you please make sure that alll of these trusses are added to `ci.yaml`?

PR #162, opened 2024-01-18
adding truss for stablelm zephyr 3b

htrivedi99:


stablelm-zephyr-3b/config.yaml
```diff
@@ -0,0 +1,15 @@
...
+requirements:
+  - accelerate==0.25.0
+  - transformers==4.35.2
+  - torch==2.1.0
+resources:
+  cpu: "2"
+  memory: 10Gi
+  use_gpu: true
+  accelerator: A10G
+system_packages: []
```
bolasim, line 15: Maybe we can increase predict_concurrency here to allow a few concurrent requests (maybe 3?)

bolasim (approved):

PR #170, opened 2024-01-26
Adding truss for dis segmentation

htrivedi99:


dis-segmentation/model/helpers.py
```diff
@@ -0,0 +1,156 @@
...
+
+# project imports
+from data_loader_cache import im_preprocess, im_reader, normalize
+from models import *
+
+
+def download_models():
+    device = "cuda" if torch.cuda.is_available() else "cpu"
+    # Download official weights
+    if not os.path.exists("./saved_models"):
```
bolasim, line 23: nit: better to use `external_data` in the config for something like this.

dis-segmentation/config.yaml
```diff
@@ -0,0 +1,20 @@
...
+external_package_dirs: []
+model_metadata:
+  example_model_input: {"input_image": "<base64 image string>"}
+model_name: DIS Segmentation
+python_version: py310
+requirements:
+  - torch==2.1.0
+  - Pillow==9.4.0
+  - numpy==1.23.5
+  - gdown==4.7.3
```
bolasim, line 11: you can drop this if you use external data

dis-segmentation/README.md
```diff
@@ -0,0 +1,104 @@
...
+The model only has one input:
+
+- `input_image` (required): An image represented as a base74 string.
+
+The output of the model contains two images in the form of a base64 string.
+Example model output:
+```json
+{
+    "img_without_bg": "Base64 string image",
+    "image_mask": "Base64 string image"
```
bolasim, line 47: is the mask optional?

bolasim (approved):

dis-segmentation/README.md
htrivedi99: yeah, the mask is optional. But with background removal tasks, the mask comes in handy

dis-segmentation/model/helpers.py
htrivedi99: I need to save the model weights inside the cloned repo. `external_data` gets executed before the clone I believe

PR #177, opened 2024-01-30
Adding flan-t5-xl-trt-llm

htrivedi99:


flan-t5-xl/flan-t5-xl-trt-llm/config.yaml
```diff
@@ -0,0 +1,21 @@
...
+  repo_id: htrivedi99/flan-t5-xl-trt-test
+model_name: Flan T5 XL TRT
+requirements:
+  - mpi4py==3.1.5
+  - pynvml==11.5.0
+resources:
+  accelerator: A100
+  use_gpu: true
+secrets:
+  hf_access_token: HF_TOKEN_HERE
```
bolasim, line 16: Can you make that repo public so that you don't need the HF token

flan-t5-xl/flan-t5-xl-trt-llm/config.yaml
```diff
@@ -0,0 +1,21 @@
...
+runtime:
```
bolasim, line 17: This really high number is usually good for continuous batching. This doesn't seem to support that right now. @joostinyi @aspctu what would be a good number here?

flan-t5-xl/flan-t5-xl-trt-llm/model/model.py
```diff
@@ -0,0 +1,74 @@
...
+
+class Model:
+    def __init__(self, **kwargs):
+        self._config = kwargs["config"]
+        self._data_dir = kwargs["data_dir"]
+        self._secrets = kwargs["secrets"]
+        self.model = None
+        self.tokenizer = None
+        self.model_config = None
+        self.hf_token = self._secrets["hf_access_token"]
```
bolasim, line 18: let's drop this and make the engine public

bolasim (commented):
Looks great. Please don't forget to the delete the old dir for flan-t5-xl so that it's not duplicated in the repo after it's moved to this subfolder.

bolasim (approved):

PR #186, opened 2024-02-06
Speculative Decoding Improvements

marius-baseten:
* Use Greedy sampling for draft model.
* Add model cache.
* Fix debugging prints.

### DIFF
--- a/speculative_decoding/config.yaml
+++ b/speculative_decoding/config.yaml
@@ -1,18 +1,21 @@
 base_image:
-  image: nvcr.io/nvidia/tritonserver:24.01-trtllm-python-py3
-  python_executable_path: /usr/bin/python3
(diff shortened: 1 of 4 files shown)


speculative_decoding/config.yaml
```diff
@@ -1,18 +1,21 @@
 base_image:
-  image: nvcr.io/nvidia/tritonserver:24.01-trtllm-python-py3
-  python_executable_path: /usr/bin/python3
+  image: "nvcr.io/nvidia/tritonserver:24.01-trtllm-python-py3"
```
aspctu, line 2: nit: Do you need quotes for this yaml spec? I think as long as the values are not to be confused with some part of the yaml syntax, it's not needed.

aspctu (approved):

PR #187, opened 2024-02-06
playground-v2 on trt

pankajroark:
TODO: Remove duplication with the sdxl-trt truss in a follow up.


aspctu (approved):

aspctu:
TODO to update README

PR #192, opened 2024-02-08
Add mixtral h100 truss

pankajroark:
mpirun was broken, I fixed that in trt-llm template in packages/client.py. Other than that only generate.yaml and the READMEs are important in this review, other code is auto generated.


joostinyi (approved):
lgtm

PR #193, opened 2024-02-09
sdxl h100 trt truss

pankajroark:
todo: Use templates to avoid duplication across various sdxl trt trusses. Let me do that in a follow-up.

### DIFF
--- a/stable-diffusion/stable-diffusion-xl-1.0-trt-h100/config.yaml
+++ b/stable-diffusion/stable-diffusion-xl-1.0-trt-h100/config.yaml
@@ -1,0 +1,65 @@
+base_image:
+  image: nvcr.io/nvidia/pytorch:23.11-py3
+  python_executable_path: /usr/bin/python
+description: Generate original images from text prompts.
+environment_variables:
+  HF_HUB_ENABLE_HF_TRANSFER: 1
+external_package_dirs: []
+model_metadata:
+  example_model_input: {"prompt": "Astronaut in a jungle, cold color palette, muted colors, detailed, 8k"}
+  pretty_name: Stable Diffusion XL - TensorRT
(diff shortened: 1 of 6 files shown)


aspctu (approved):
We should figure out the right story with templates in truss-examples and the templates inside of truss. Can be done off-line / follow-up.

stable-diffusion/stable-diffusion-xl-1.0-trt-h100/config.yaml
```diff
@@ -0,0 +1,65 @@
...
+    allow_patterns:
+      - "*.json"
+      - "*.fp16.safetensors"
+      - sd_xl_base_1.0.safetensors
+  - repo_id: stabilityai/stable-diffusion-xl-refiner-1.0
+    allow_patterns:
+      - "*.json"
+      - "*.fp16.safetensors"
+      - sd_xl_refiner_1.0.safetensors
+python_version: py39
```
joostinyi, line 34: let's remove this – we use base image feature anyway

stable-diffusion/stable-diffusion-xl-1.0-trt-h100/config.yaml
```diff
@@ -0,0 +1,65 @@
...
+system_packages:
+- python3.10-venv
```
joostinyi, line 36: maybe this should be 3.11 too

joostinyi (approved):

stable-diffusion/stable-diffusion-xl-1.0-trt-h100/config.yaml
pankajroark: tensorrt seems to want 3.10

PR #194, opened 2024-02-09
Speculative Decoding Updates & Fixes

marius-baseten:
* Replace baseline mistral with instruction tuned mistral as target model.
* Parameterize the number of draft tokens used and add conventional generation as fallback for no draft tokens (streaming API is supported, only yields full output in one chunk).
* Address some TODOs.
* Fix type annotations.

### DIFF
--- a/speculative_decoding/engine_building/build_gpt2.py
+++ b/speculative_decoding/engine_building/build_gpt2.py
@@ -12,7 +12,6 @@ HF_DIR = f"/root/workbench/{MODEL_NAME}_hf"
 FT_DIR = f"/root/workbench/{MODEL_NAME}_ft"
 ENGINE_DIR = f"/root/workbench/{MODEL_NAME}_engine"
 
-MAX_DRAFT_LEN = 4
 
 snapshot_download(
     MODEL_HF_ID,
@@ -47,10 +46,11 @@ command = [
     "--use_inflight_batching",
     "--use_gemm_plugin=float16",
     "--multi_block_mode",
-    "--max_batch_size=16",
-    "--max_input_len=512",
+    "--max_batch_size=32",
+    "--max_input_len=1024",
+    "--max_output_len=2048",
     "--use_paged_context_fmha",
-    "--gather_all_token_logits",
+    # "--gather_all_token_logits",
     f"--output_dir={ENGINE_DIR}",
 ]
 print(" ".join(command))
--- a/speculative_decoding/engine_building/build_gpt2.py
+++ b/speculative_decoding/engine_building/build_gpt2.py
@@ -50,7 +50,6 @@ command = [
     "--max_input_len=1024",
     "--max_output_len=2048",
     "--use_paged_context_fmha",
-    # "--gather_all_token_logits",
     f"--output_dir={ENGINE_DIR}",
 ]
 print(" ".join(command))
--- a/speculative_decoding/model/model.py
+++ b/speculative_decoding/model/model.py
@@ -1,12 +1,13 @@
 """
-This model downloads a target and draft model engine, gets their corresponding
-tokenizers from HF and starts a tritonserver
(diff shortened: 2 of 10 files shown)


speculative_decoding/model/model.py
```diff
@@ -47,40 +48,42 @@ def __init__(self, **kwargs):
...
     def load(self):
-        tensor_parallel_count = self._config["model_metadata"].get(
-            "tensor_parallelism", 1
-        )
-        pipeline_parallel_count = self._config["model_metadata"].get(
-            "pipeline_parallelism", 1
-        )
+        if not helpers.is_triton_server_alive():
+            tensor_parallel_count = self._config["model_metadata"].get(
+                "tensor_parallelism", 1
```
aspctu, line 53: nit: Let's be consistent re: what we call these variables and the structure for the `config.yaml` so that they're all the same across Truss'. I started on this with the template work, we can tackle this in a follow-up.

speculative_decoding/engine_building/build_gpt2.py
```diff
@@ -47,10 +46,11 @@
...
     "--use_gemm_plugin=float16",
     "--multi_block_mode",
-    "--max_batch_size=16",
-    "--max_input_len=512",
+    "--max_batch_size=32",
+    "--max_input_len=1024",
+    "--max_output_len=2048",
     "--use_paged_context_fmha",
-    "--gather_all_token_logits",
+    # "--gather_all_token_logits",
```
pankajroark, line 53: Generally, we try to delete instead of comment code.

speculative_decoding/model/model.py
```diff
@@ -47,43 +48,45 @@ def __init__(self, **kwargs):
...
-            use_auth_token=hf_access_token,
-        )
+            env = {}
+            try:  # Secrets does not have __contains__.
+                hf_access_token = self._secrets["hf_access_token"]
+                env["HUGGING_FACE_HUB_TOKEN"] = hf_access_token
+            except (
+                Exception
+            ):  # `SecretNotFound` from truss template cannot be referenced.
+                hf_access_token = None
```
pankajroark, line 66: May be extract out into a function:
`def supplied_hf_token(self) -> Optional[str]:`

speculative_decoding/model/model.py
```diff
@@ -47,43 +48,45 @@ def __init__(self, **kwargs):
...
+
+            # Target model.
+            huggingface_hub.snapshot_download(
+                self._config["model_metadata"]["engine_repository"],
+                local_dir=os.path.join(TRITON_DIR, TARGET_MODEL_KEY, "1"),
+                local_dir_use_symlinks=False,
+                max_workers=4,
+                use_auth_token=hf_access_token,
+            )
```
pankajroark, line 75: nit: this and statement below are too similar, perhaps refactor.

speculative_decoding/packages/helpers.py
```diff
@@ -66,7 +68,7 @@ class _TimingManager(object):
     def __init__(self):
         self._timings = collections.defaultdict(_Timing)
         self._order = []
-        self._total = None
+        self.total_section_name = None
```
pankajroark, line 71: nit: Is this meant to be public? We generally keep all member variable private and create a public property function to expose.

speculative_decoding/packages/triton_model_repo/target_model/config.pbtxt
```diff
@@ -219,7 +219,7 @@ parameters: {
 parameters: {
   key: "batch_scheduler_policy"
   value: {
-    string_value: "max_utilization"
+    string_value: "guaranteed_no_evict"
```
pankajroark, line 222: Is this temporarily for debugging for meant to be permanent? A comment explanation will be useful here.

pankajroark (approved):

PR #197, opened 2024-02-13
Limit Parallelism of CI Job to 3

bolasim:


squidarth (approved):
lgtm (ran it here: Ran it here: https://github.com/basetenlabs/truss-examples/actions/runs/7885542880, looks good). I think we could likely dial this up a little higher, but there's no time pressure on this job so I think it's fien

jrochette (approved):

PR #200, opened 2024-02-13
Add NaturalSQL-7b & PhotoMaker Truss Examples

enoreyes:
This is a demonstration of Factory's code-generation capabilities following our meeting @htrivedi99 @bolasim. Here are the linear issues used for these:

## Create a new truss example with TencentARC/PhotoMaker
https://github.com/TencentARC/PhotoMaker

We need to create a new truss example. It should be in the top level directory as a new folder. It should contain a config and a readme, with a model folder containing an empty init and a model.py

The model.py should include the code for loading the PhotoMakerStableDiffusionXLPipeline pipeline and running predict. You can base this on the other model.py files, but use the logic from the photomaker_demo.ipynb for specific details on running inference. Importantly, this is an image to image model - so it should be modeled after the other image to image models in the repository.

This one has system dependencies and a big model, so include the system dependencies and cache the photomaker huggingface model in the config.

## Create a new truss example with natural-sql-7b
https://github.com/cfahlgren1/natural-sql

We need to create a new truss example. It should be in the top level directory as a new folder. It should contain a config and a readme, with a model folder containing an empty init and a [model.py](http://model.py/)

The model.py should include the code for loading the chatdb/natural-sql-7b model, and then running predict on this model. You can base this on the other model.py files, but use the logic from natural-sql-7b.ipynb for specific details on running inference and loading the model.


photomaker/config.yaml
```diff
@@ -0,0 +1,32 @@
...
+hf_model_cache:
+  function: hf_hub_download
+  parameters:
+    repo_id: "TencentARC/PhotoMaker"
+    filename: "photomaker-v1.bin"
+
+# GPU Configuration
+gpu_requirements:
+  memory_min_gb: 11
+  data_type_for_non_bfloat16_gpus: "torch.float16"
```
bolasim, line 32: none of this is valid syntax for the config.yaml

photomaker/model/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""
+This __init__.py file initializes the model directory as a Python package for the PhotoMaker Truss example.
+"""
+
+# Import necessary modules from the package here if needed
+# from .model import PhotoMakerStableDiffusionXLPipeline
```
bolasim, line 6: this file is generally empty. You can also just drop it

photomaker/model/model.py
```diff
@@ -0,0 +1,67 @@
+import numpy as np
+import torch
+from huggingface_hub import hf_hub_download
+from photomaker import PhotoMakerStableDiffusionXLPipeline
+from PIL import Image
+
+
+class PhotoMakerModel:
```
bolasim, line 8: must be called `Model` and contain `load` and `predict` method

photomaker/model/model.py
```diff
@@ -0,0 +1,67 @@
...
+    def __init__(self):
+        self.model = None
+        self.device = "cuda" if torch.cuda.is_available() else "cpu"
+
+    def load_model(self):
+        # Download the model from Hugging Face Hub
+        model_path = hf_hub_download(
+            repo_id="TencentARC/PhotoMaker",
```
bolasim, line 16: you should use `model_cache` instead of downloading the model here. Take a look at some of the other examples please.

photomaker/model/model.py
```diff
@@ -0,0 +1,67 @@
...
+            filename="photomaker-v1.bin",
+            repo_type="model",
+        )
+        self.model = PhotoMakerStableDiffusionXLPipeline.from_pretrained(model_path).to(
```
bolasim, line 20: i think you can `from_pretrained(model_path, device=self.device)` instead which might be faster. Can you try it?

photomaker/model/model.py
```diff
@@ -0,0 +1,67 @@
...
+            self.device
+        )
+
+    def predict(self, input_image: Image) -> Image:
+        """
+        Takes an input image and produces an output image using the PhotoMakerStableDiffusionXLPipeline model.
+        """
+        # Ensure the model is loaded
+        if self.model is None:
```
bolasim, line 29: should never load in the predict method unless explicity/intentionally hotswapping weights

photomaker/model/model.py
```diff
@@ -0,0 +1,67 @@
...
+        model_path = hf_hub_download(
+            repo_id="TencentARC/PhotoMaker",
+            filename="photomaker-v1.bin",
+            repo_type="model",
+        )
+        self.model = PhotoMakerStableDiffusionXLPipeline.from_pretrained(model_path).to(
+            self.device
+        )
+
+    def predict(self, input_image: Image) -> Image:
```
bolasim, line 24: input type has to json serialiazable unless handled by the pre-process funtion (whose input also has to be json serializable).

Output also has to be json serializable.

photomaker/model/model.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        Converts a tensor to a PIL Image.
+        """
+        tensor = tensor.squeeze(0).permute(1, 2, 0).cpu().numpy()
+        tensor = np.clip(tensor, 0, 255).astype(np.uint8)
+        image = Image.fromarray(tensor)
+        return image
+
+
+# Example usage
```
bolasim, line 62: no mains in model.py

bolasim (changes requested):
Great start. Looking forward to seeing the improvements. Feel free to make a baseten account and try deploying these to ensure everything works end to end.

photomaker/model/model.py
htrivedi99: Also, I'm pretty sure `PhotoMakerStableDiffusionXLPipeline` is not a real class.

natural-sql-7b/README.md
```diff
@@ -0,0 +1,65 @@
...
+The `natural-sql-7b` model showcases exceptional ability in generating correct SQL queries on novel schemas. It is benchmarked using the SQL-Eval framework and excels in understanding complex query intentions.
+
+## Setup and Dependencies
+
+To use the `natural-sql-7b` model, ensure you have the following dependencies installed:
+
+```bash
+pip install transformers==4.35.2 accelerate sqlparse
+```
+
```
htrivedi99, line 16: Doesn't talk about using a truss at all. It should say something like "cd into truss-examples/natural-sql-7b", then run "truss push".

natural-sql-7b/config.yaml
```diff
@@ -0,0 +1,25 @@
+environment_variables: {}
+external_package_dirs: []
+model_framework: transformers
+model_type: AutoModelForCausalLM
+description: NaturalSQL-7B is a state-of-the-art model for generating SQL queries from natural language questions, developed by ChatDB.
+model_metadata:
+  avatar_url: https://cdn-uploads.huggingface.co/production/uploads/648a374f00f7a3374ee64b99/5ynfoKPzI3_-WasQQt7qR.png
+  cover_image_url: https://global-uploads.webflow.com/6348b2d49808811e3f7a0fff/640690727b722a05771960ec_graphic-data-p-800.png
```
htrivedi99, line 8: The image here are just random images scraped from hugging face or github. They should be proper logos or images representing the creator of the model.

natural-sql-7b/config.yaml
```diff
@@ -0,0 +1,25 @@
...
+  tags:
+  - text-to-SQL
+  - natural language processing
+  - SQL generation
```
htrivedi99, line 12: These tags are made up. For example our [nsql](https://github.com/basetenlabs/truss-examples/blob/main/nsql/config.yaml) truss has only one tag `code-generation`

natural-sql-7b/model/model.py
```diff
@@ -0,0 +1,58 @@
...
+            torch_dtype=torch.float16,
+        )
+
+    def predict(self, questions):
+        """
+        Generates SQL queries from a list of questions.
+
+        :param questions: List of natural language questions.
+        :return: List of SQL queries corresponding to the input questions.
+        """
```
htrivedi99, line 27: I like that its documenting each function. It can help our users understand the code a bit quicker.

htrivedi99 (changes requested):
There are a few problems with the overall layout of each truss.

photomaker/model/model.py
enoreyes: The photomaker repository instantiates the pipeline this way (`PhotoMakerStableDiffusionXLPipeline` is defined in the photomaker library). I haven't tested performance in either case however.

PR #201, opened 2024-02-14
adding llava sgl

htrivedi99:


llava/llava-1.6-sgl/model/model.py
```diff
@@ -0,0 +1,101 @@
...
+import sglang as sgl
+from PIL import Image
+
+BASE64_PREAMBLE = "data:image/png;base64,"
+
+
+class Model:
+    def __init__(self, **kwargs):
+        self.tokenizer = None
+        self.runtime = sgl.Runtime(
```
bolasim, line 16: you can actually move this to load unless it doesn't work there.

bolasim (approved):

PR #205, opened 2024-02-18
Add metavoice base model

vshulman:
https://github.com/metavoiceio/metavoice-src

see note re: `flash_attn` in README


metavoice-1b/config.yaml
```diff
@@ -0,0 +1,29 @@
...
+    allow_patterns:
+      - mbd_comp_8.pt
+  - repo_id: facebook/encodec_24khz
+    allow_patterns:
+      - "*.safetensors"
+
+requirements_file: ./requirements.txt
+resources:
+  accelerator: "A100"
+  cpu: '2'
```
squidarth, line 23: i would leave out the cpu & memory requiremetns

metavoice-1b/model/model.py
```diff
@@ -0,0 +1,147 @@
...
+    """Data class for holding model state."""
+    spkemb_model: Optional[torch.nn.Module]
+    first_stage_model: Optional[Model]
+    second_stage_model: Optional[Model]
+    enhancer: Optional[object]
+
+class Model:
+    def __init__(self, **kwargs):
+        self._secrets = kwargs["secrets"]
+        self.config = ServingConfig()
```
squidarth, line 64: could we call this `DefaultInferenceParams` or something like that instead? I feel like there are too many things called `config` right now (the `InferenceConfig`, `config.yaml`, and this)

squidarth (approved):
lgtm! as an FYI -- we are soon going to have a way to write trusses w/ pydantic type input/output types so you can use your `TTSRequest` for that instead of building the object later.

vshulman:
@squidarth lmk if this looks better now

PR #206, opened 2024-02-20
Github action for deploying new library models.

squidarth:
# Summary

Adds a github action that runs on merges to main that looks at what files have changed. If any of these files match any of the paths in the model library, then, upsert the new library model.

# Testing

[TODO] 

I added a way to trigger this workflow with a custom git diff string. For instance, if you trigger this with: "e7a81ae..0c78c6f", it will use that as the diff. This makes it easier to test.

Before merging this PR, I'll try triggering this workflow using the GH CLI.

# Follow-ups

- Add command to push models to production

### DIFF
--- a/.github/workflows/deploy-library-model.yml
+++ b/.github/workflows/deploy-library-model.yml
@@ -1,0 +1,58 @@
+name: Deploy Library Model
+
+on:
+  schedule:
+    # Runs daily at 6am UTC
+    - cron:  '0 6 * * *'
+  workflow_dispatch:
+    inputs:
+      diff:
+        type: string
+
+jobs:
+  detect_changes:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python environment
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.11.4'
+      - name: Install dependencies (if any)
+        run: |
+          python -m pip install --upgrade pip
+          pip install git+https://github.com/basetenlabs/truss.git requests tenacity --upgrade
+      - name: Run the detect changes script
+        run: |
+          LIBRARY_MODELS_TO_DEPLOY=python ./bin/detect_library_changes.py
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy-library-model.yml
```diff
@@ -0,0 +1,58 @@
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install git+https://github.com/basetenlabs/truss.git requests pyyaml tenacity --upgrade
+      - name: Test => (${{ matrix.library_model_to_deploy }}
+        run: |
+          python ./bin/test_example.py ${{secrets.BASETEN_API_KEY}} ${{matrix.library_model_to_deploy}}
+      - name: Deploy => (${{ matrix.library_model_to_deploy }} to staging
+        run: |
+          echo "Deploying ${{ matrix.library_model_to_deploy}} to staging"
+          python ./bin/upsert_library_model.py ${{secrets.UPSERT_LIBRARY_MODELS_STAGING_API_KEY}} ${{matrix.library_model_to_deploy}} $GITHUB_SHA https://app.staging.baseten.co
```
jrochette, line 58: wondering if we should use env vars instead here? Using this like that will result in the API being printed in the github action output logs right?

bin/upsert_library_model.py
```diff
@@ -0,0 +1,36 @@
...
+import sys
+
+import requests
+import yaml
+from tenacity import retry, stop_after_attempt, wait_fixed
+
+
+@retry(wait=wait_fixed(20), stop=stop_after_attempt(3), reraise=True)
+def upsert_pretrained_model(api_key, model_slug, commit_sha, baseten_url):
+    url = f"{baseten_url}/upsert_pretrained_model"
```
jrochette, line 11: nit: maybe it would be worth logging / printing some here? Something like `Upserting library model {slug} with commit {commit_sha}`

jrochette (approved):
Looks good. Main concern is about leaking the API key in the logs

PR #208, opened 2024-02-21
Adding TRT-LLM Gemma 7B engine

aspctu:
**do not merge** but feel free to review


gemma/gemma-7b-it-trtllm/packages/client.py
```diff
@@ -0,0 +1,155 @@
...
+    def start_server(
+        self,
+        mpi: int = 1,
+        env: dict = {},
+    ):
+        """Triton Inference Server has different startup commands depending on
+        whether it is running in a TP=1 or TP>1 configuration. This function
+        starts the server with the appropriate command."""
+        if mpi == 1:
+            command = [
```
squidarth, line 71: instead of manually building up this list of strings iteratively, a more readable/scalable approach here could be adding a level of indirection, and instead builiding up some pydantic/dataclass object. Then you could have a function that converts that object into the bash command to run.

that way this TritonClient code does not have to concern itself with formatting strings correctly/getting argument order correct.

squidarth:
would be possible to move some of the triton client stuff into the truss repo (under truss-utils), so we don't have to copy/paste in a bunch places?

gemma/gemma-7b-it-trtllm/data/.gitattributes
marius-baseten, line 1: Could it be that this whole directory is a spurious leftover and not needed? I remember looking into this for other trusses and not figuring out any need for it. - Remove it if possible (ideally also for other trusses).

And in case it's needed for anything, replace the attributes file with an empty `.gitkeep`.

PR #211, opened 2024-02-23
Adding fix for faster-whisper

htrivedi99:


squidarth (approved):
great, thanks for removing the old base image thing too

squidarth:
Also @htrivedi99 , just to confirm, you tested this right?

htrivedi99:
> Also @htrivedi99 , just to confirm, you tested this right?

yes

PR #212, opened 2024-02-24
Cleanup config to only include minimal number of keys

bolasim:
Just ran this very minimal cleanup script. We could all the linting at some point but didn't want to dump too much time on figuring that out right now.

Excluded the education trusses 01-10

```
import pathlib
import yaml
from truss.truss_config import TrussConfig

def cleanup_config_files(start_path='.'):
    start_path = pathlib.Path(start_path)
    for path in start_path.rglob('*'):
        if path.is_dir():
            config_file = path / 'config.yaml'
            if config_file.exists():
                config_d = yaml.safe_load(config_file.read_text())
                config_file.write_text(yaml.safe_dump(TrussConfig(**config_d).to_dict(verbose=False)))

cleanup_config_files()
```

## Testing:
Ran [passing CI](https://github.com/basetenlabs/truss-examples/actions/runs/8031945180)


squidarth (approved):
yay thanks for doing this @bolasim ! didn't review super closely, but feel ok given CI passing 👍

PR #218, opened 2024-02-27
Making faster whisper input more uniform

htrivedi99:


whisper/faster-whisper-v2/model/model.py
```diff
@@ -16,13 +17,39 @@ def load(self):
...
     def preprocess(self, request: Dict) -> Dict:
-        resp = requests.get(request["url"])
-        return {"response": resp.content}
+        audio_base64 = request.get("audio")
+        url = request.get("url")
+
+        if audio_base64 and url:
+            return {
+                "status": 400,
+                "msg": "Only a base64 audio file OR a URL can be passed to the API, not both of them!",
```
vshulman, line 26: nit: I think we should avoid exclamation points in these. Maybe something a little more neutral like:
"message": "Please provide either a base64-encoded audio file or a URL, but not both."

there may be enough here to warrant wrapping this in a little helper method that can be included in other models that take audio files as input

vshulman (changes requested):
Suggestion re: language and factoring out some of the code into its own method

bolasim (commented):
the source for both of these is templates from `templates/fasater-whisper`. maybe edit there and just regenerate these. 

That being said, I don't like these status codes. it'd be great if you can set them in the truss response (we'll add that soon), but it won't make send now because the HTTP status code will always be a 200

squidarth:
> That being said, I don't like these status codes. it'd be great if you can set them in the truss response (we'll add that soon), but it won't make send now because the HTTP status code will always be a 200

fwiw, once we have truss types fully out, we can have this be a pydantic class w/ a validation, and return nice 400 errors.

bolasim (approved):

vshulman (approved):

PR #222, opened 2024-02-28
flat t5 trtllm follow up

pankajroark:
* Stop early -- added a stopping criteria for this, couldn't find another way
* Print inference time
* Use hf_access_token from secrets


aspctu (approved):
Lgtm, I wonder if there's a way we can pass the EOS token ID as part of the request to signify to TRT-LLM to stop on EOS. We do something similar with the Triton server implementation.

PR #225, opened 2024-02-29
Whisper Torchserve truss

htrivedi99:


whisper/whisper-torchserve/model/model.py
```diff
@@ -0,0 +1,56 @@
...
+                "--foreground",
+                "--no-config-snapshots",
+                "--ts-config",
+                f"{self._data_dir}/config.properties",
+            ],
+            check=True,
+        )
+
+    def load(self):
+        snapshot_download(
```
vshulman, line 37: nit: any reason not to cache?

whisper/whisper-torchserve/config.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - httpx==0.27.0
+resources:
+  accelerator: T4
+  use_gpu: true
+secrets: {}
+system_packages:
+  - ffmpeg
+  - openjdk-11-jdk
+runtime:
+  predict_concurrency: 128
```
vshulman, line 21: @htrivedi99 you feel good about this number? higher than I would expect; e.g. if the batch size is 4 in torchserver, then i think this means we'll have 31 batches (128/4 - 1) waiting. But i may be thinking about it wrong.

vshulman (approved):

whisper/whisper-torchserve/README.md
```diff
@@ -0,0 +1,56 @@
...
+import base64
+
+def wav_to_base64(file_path):
+    with open(file_path, "rb") as wav_file:
+        binary_data = wav_file.read()
+        base64_data = base64.b64encode(binary_data)
+        base64_string = base64_data.decode("utf-8")
+        return base64_string
+
+resp = requests.post(
```
bolasim, line 43: would be cool to send multiple requests in parallel to show dynamic batchung.

whisper/whisper-torchserve/config.yaml
bolasim: I think it's probably okay to leave it be quite large (since truss is just proxying). the desired number of pending batches can be controlled in the deployment autoscaling settings using the target concurrency

whisper/whisper-torchserve/data/config.properties
```diff
@@ -0,0 +1,20 @@
...
+async_logging=true
+
+models={\
+  "whisper_base": {\
+    "1.0": {\
+        "defaultVersion": true,\
+        "marName": "whisper_base.mar",\
+        "minWorkers": 1,\
+        "maxWorkers": 2,\
+        "batchSize": 4,\
```
bolasim, line 13: would be good to add comments explaining what these are?

whisper/whisper-torchserve/data/config.properties
```diff
@@ -0,0 +1,20 @@
+inference_address=http://0.0.0.0:8888
+batch_size=4
+ipex_enable=true
+async_logging=true
+
+models={\
+  "whisper_base": {\
+    "1.0": {\
+        "defaultVersion": true,\
+        "marName": "whisper_base.mar",\
```
bolasim, line 10: would be good to add comments explaining where this is stored or expected?

whisper/whisper-torchserve/model/model.py
```diff
@@ -0,0 +1,56 @@
...
+            "htrivedi99/whisper-torchserve",
+            local_dir=os.path.join(self._data_dir, "model_store"),
+            max_workers=4,
+        )
+        print("Downloaded weights succesfully!")
```
bolasim, line 42: let's using `logging.info` or `logging.debug` instead of print

whisper/whisper-torchserve/model/model.py
```diff
@@ -0,0 +1,56 @@
...
+        self._data_dir = kwargs["data_dir"]
+        self._model = None
+
+    def start_tochserver(self):
+        subprocess.run(
+            [
+                "torchserve",
+                "--start",
+                "--model-store",
+                f"{self._data_dir}/model_store",
```
bolasim, line 25: This doesn't look like it's checked in. Do you need to provide a simple script to generate it or do we add it to `external_data` in the config so its downloaded?

bolasim (changes requested):

bolasim (approved):

PR #228, opened 2024-03-03
Update Mistral Instruct v0.1 to v0.2

vshulman:


mistral/mistral-7b-instruct/model/model.py
```diff
@@ -16,13 +16,13 @@ def __init__(self, **kwargs):
 
     def load(self):
         self.model = AutoModelForCausalLM.from_pretrained(
-            "mistralai/Mistral-7B-Instruct-v0.1",
+            "mistralai/Mistral-7B-Instruct-v0.2",
```
bolasim, line 19: consider using `self._config["model_cache"][0]["repo_id"] instead here and below so that the model id only lives in a single place.

bolasim (approved):

mistral/mistral-7b-instruct/model/model.py
vshulman: good idea -- fixed.

PR #229, opened 2024-03-04
Fixing issue with torchserve not being available on startup

htrivedi99:


whisper/whisper-torchserve/data/config.properties
```diff
@@ -9,10 +9,10 @@ models={\
         "defaultVersion": true,\
         "marName": "whisper_base.mar",\
         "minWorkers": 1,\
-        "maxWorkers": 2,\
-        "batchSize": 4,\
-        "maxBatchDelay": 500,\
-        "responseTimeout": 24\
+        "maxWorkers": 4,\
+        "batchSize": 16,\
+        "maxBatchDelay": 1000,\
```
bolasim, line 14: 1 second might be too high of a latency hit here. I'd something lik 250ms and leave a comment that explains what this is.

PR #234, opened 2024-03-11
Fix method spelling

msfeldstein:
s/tochserver/torchserver


htrivedi99 (approved):
Thank you for catching this!

PR #242, opened 2024-03-21
Adding deepspeed to xtts-streaming

htrivedi99:


xtts-streaming/model/model.py
```diff
@@ -2,6 +2,7 @@
 import io
 import logging
 import os
+import time
```
bolasim, line 5: I don't think you need this.

bolasim (approved):
Can you please add the Dockerfile for the base image in this directory?

PR #258, opened 2024-04-10
Update typo in model.py "top_p" -> "top_k"

IanCarrasco:
There appears to be a typo when setting the parameters for Mistral inference. top_k is being set using the top_p entry in the request dict


squidarth (approved):

squidarth:
@IanCarrasco thanks for catching/fixing that!

PR #272, opened 2024-04-19
Adding files for mixtral 8x22b instruct trt int8 quantized

htrivedi99:


mistral/mixtral-8x22b-trt-int8-weights-only/README.md
```diff
@@ -0,0 +1,54 @@
+# Mixtral-8x7B-Instruct Truss
+
+This is a [Truss](https://truss.baseten.co/) for Mixtral 8x7B Instruct. This README will walk you through how to deploy this Truss on Baseten to get your own instance of Mistral 7B Instruct.
```
vshulman, line 3: version needs updating (and later in doc)

vshulman (changes requested):

vshulman (approved):
approved, let's just add support for the hf token

PR #287, opened 2024-05-03
fix issue with nomic model

philipkiely-baseten:
Fixes Nomic Embed V1.5. Context https://huggingface.co/nomic-ai/nomic-embed-text-v1.5/discussions/19


nomic/nomic-embed-v1_5/model/model.py
```diff
@@ -1,10 +1,12 @@
 import torch.nn.functional as F
+from huggingface_hub import login as hf_login
```
joostinyi, line 2: would it be useful to explicitly pin this package in our requirements?

joostinyi (approved):

PR #294, opened 2024-05-15
Fixed deploying whisper streaming to baseten by removing python executable path in config file

usman61:
I was trying to deploy whisper streaming on baseten but I was getting error Invalid relative python executable path /usr/bin/python3 in config.yaml file.
after removing the model deployed successfully to baseten.


squidarth:
@usman61 This should be fixed in Truss 0.9.12, could you try that and let us know if it works for you?

If not, please reopen this

usman61:
@squidarth I tried by upgrading truss version to 0.9.12 but it isn't working as it has some package dependency issues as shown in given figure. I don't have access to reopen the pull request. could you please reopen it and merged with the main.

image

PR #304, opened 2024-06-06
Lora example

pankajroark:
Allows loading Lora dynamically as part of request. Lora are cached and reused. On cache eviction Lora are loading again as needed. Currently Lora is only accepted via a HuggingFace repo path.


lora/README.md
```diff
@@ -0,0 +1,3 @@
+# LoRA
```
aspctu, line 1: Would be good to have the command to generate an engine with the appropriate flags for LoRA

lora/config.yaml
```diff
@@ -0,0 +1,40 @@
...
+live_reload: false
+model_class_filename: model.py
+model_class_name: Model
+model_framework: custom
+model_metadata:
+  tags:
+  - text-generation
+  - openai-compatible
+model_module_dir: model
+model_name: Mistral 7B Instruct TRT-LLM no LoRA
```
joostinyi, line 28:
```suggestion
model_name: Mistral 7B Instruct TRT-LLM with LoRA 
```

lora/README.md
joostinyi: I think we can link https://github.com/triton-inference-server/tensorrtllm_backend/blob/b93b948be37046cf2a8a222067771fb366baf2a8/inflight_batcher_llm/README.md#running-lora-inference-with-inflight-batching

joostinyi (approved):

PR #311, opened 2024-06-25
Updating ComfyUI readme to include build commands

htrivedi99:


comfyui-truss/README.md
```diff
@@ -30,22 +30,43 @@ git clone https://github.com/basetenlabs/truss-examples/
...
+  use_gpu: true
+secrets: {}
+system_packages:
+  - wget
+  - ffmpeg
+  - libgl1-mesa-glx
 ```
 
-In this case, I have 2 models: SDXL and a ControlNet. Each model needs to have 2 things, `url` and `path`. The `url` is the location for downloading the model. The `path` is where this model will get stored inside the Truss. For the path, follow the same guidelines as used in ComfyUI. Models should get stored inside `checkpoints`, ControlNets should be stored inside `controlnet`, etc.
+Here is a breakdown of the things happenning in `build_commands`. First, we are cloning the ComfyUI repository, checking out a specific commit, and installing the required python packages to run Comfy. Next, we are use the `cd` commmand to go into the `custom_nodes` directory and then cloning and installing requirements for each custom node. Similarly for the checkpoints, we use the `wget` utility to download the checkpoints and place them in the appropriate directories within ComfyUI.
```
vshulman, line 67: nit: maybe actions instead of things

vshulman: nit: Comfy should be ComfyUI

vshulman: maybe
"Next, we use the `cd` command to ensure all custom nodes are downloaded within the `custom_nodes` directory, and their respective requirements are installed"

vshulman (approved):
lgtm with minor suggestions

PR #313, opened 2024-06-26
Add warm up chains github action.

squidarth:
Adds a github action that deploys the poems.py thing every 20 mins.

Once this is working properly, we can add the remaining docs examples.


chains-examples/docs-examples/poems.py
marius-baseten, line 1: For a better example cleanup (so that I can also move the other things straight) should we use `chains-examples/docs/poems/poems.py` (and then likewise `chains-examples/docs/rag/...` etc.)?

marius-baseten (approved):

chains-examples/docs-examples/poems.py
squidarth: Yeah that sounds good

PR #317, opened 2024-07-01
Adding vllm speculative decoding example

htrivedi99:


vllm-speculative-decoding/config.yaml
```diff
@@ -0,0 +1,23 @@
+base_image:
+  image: nvcr.io/nvidia/pytorch:23.11-py3
+  python_executable_path: /usr/bin/python3
```
bolasim, line 3: just cuious why we need this base image? Can you add a cooment?

vllm-speculative-decoding/config.yaml
```diff
@@ -0,0 +1,23 @@
...
+environment_variables:
+  HF_TOKEN: ""
+external_package_dirs: []
+model_metadata:
+  main_model: meta-llama/Meta-Llama-3-8B-Instruct
+  assistant_model: ibm-fms/llama3-8b-accelerator
+  tensor_parallel: 1
+  max_num_seqs: 16
+model_name: vLLM Speculative Decoding
+python_version: py310
```
bolasim, line 14: drop python version if using base image

vllm-speculative-decoding/config.yaml
```diff
@@ -0,0 +1,23 @@
+base_image:
+  image: nvcr.io/nvidia/pytorch:23.11-py3
+  python_executable_path: /usr/bin/python3
+build_commands: []
+environment_variables:
+  HF_TOKEN: ""
```
bolasim, line 6: why here over secrets?

htrivedi99: vLLM reads only this specific environment variable for the access token. It doesn't work with secrets

vllm-speculative-decoding/config.yaml
htrivedi99: Without this base image the build does not succeed. The baseten base image does not have `nvcc`, which is required for the developer build of vLLM.

PR #318, opened 2024-07-02
added gemma2 9b and 27b vllm with streaming

dsingal0:


dsingal0:
vllm's next release will add support for gemma2 9/27B. Until then you'd have to build from source on top of a pytorch image which takes 30+ minutes to deploy. https://github.com/vllm-project/vllm/issues/5806

gemma2/gemma2-27b-it-vllm/model/model.py
```diff
@@ -0,0 +1,84 @@
...
+            )
+            print(result.stdout)
+        except subprocess.CalledProcessError as e:
+            print(f"Command failed with code {e.returncode}: {e.stderr}")
+        model_metadata = self._config["model_metadata"]
+        logger.info(f"main model: {model_metadata['main_model']}")
+        logger.info(f"tensor parallelism: {model_metadata['tensor_parallel']}")
+        logger.info(f"max num seqs: {model_metadata['max_num_seqs']}")
+
+        self.model_args = AsyncEngineArgs(
```
vshulman, line 44: potential improvement is to move everything to config, e.g. as in this example: 
https://github.com/vshulman/truss-examples/tree/main/ultravox-vllm

gemma2/gemma2-27b-it/model/model.py
```diff
@@ -0,0 +1,78 @@
...
+                "pad_token_id": self.tokenizer.eos_token_id,
+                "max_new_tokens": 1000,
+                "streamer": streamer,
+            }
+            # Spawn a thread to run the generation, so that it does not block the main
+            # thread.
+            thread = Thread(target=self.model.generate, kwargs=generation_kwargs)
+            thread.start()
+
+            def inner():
```
vshulman, line 73: could we add support for Stream = False?

gemma2/gemma2-27b-it/config.yaml
```diff
@@ -0,0 +1,26 @@
...
+- sentencepiece==0.2.0
+- tokenizers==0.19.1
+- torch==2.3.1
+- transformers==4.42.3
+- triton==2.3.1
+secrets:
+  hf_access_token: null
+# Note that we need an A100 to run this model.
+resources:
+  cpu: "8"
```
vshulman, line 23: probably okay to leave off CPU and memory

vshulman (commented):
Left a few small suggestions, but looking forward to having 27b, folks are looking forward to it.

gemma2/gemma2-27b-it-vllm/model/model.py
vshulman: we can merge without this change as other vLLM examples also support a partial list of arguments

dsingal0: Looking through it, it looks like that example uses the vllm openai server instead of explicitly instantiating the vllm AsyncLLMEngine for the model.

gemma2/gemma2-27b-it/model/model.py
dsingal0: Added. when passing stream parameter as string "False" it returns entire output.

gemma2/gemma2-27b-it/config.yaml
dsingal0: done

gemma2/gemma2-27b-it-vllm/model/model.py
```diff
@@ -0,0 +1,84 @@
...
+        self._config = kwargs["config"]
+        self.model = None
+        self.llm_engine = None
+        self.model_args = None
+        self.hf_secret_token = kwargs["secrets"]["hf_access_token"]
+        os.environ["HF_TOKEN"] = self.hf_secret_token
+        print(
+            "logging in with huggingface authentication token: ", self.hf_secret_token
+        )
+        huggingface_hub.login(token=self.hf_secret_token, add_to_git_credential=True)
```
vshulman, line 25: I don't think we need this hub login bit, setting `HF_TOKEN` should be enough (and I think you can then remove `huggingface_hub` altogether from requirements)

gemma2/gemma2-27b-it-vllm/model/model.py
vshulman: 100% -- I just think the same kwargs pattern can apply here. the benefit I see is that going forward all it would take to pass a new argument into vLLM, either the standalone OpenAI server or the Python API above, is adding it to the config.yaml.

vshulman (commented):
Do you think this pattern is expected? It looks like the instruct response completes part of the request (! 😀), which makes me think there is potentially a templating issue?

image

gemma2/gemma2-27b-it-vllm/model/model.py
dsingal0: I tried it without, because we could remove huggingface_hub and git from the requirements, but had issues with vLLM being unable to get the model. Also, by reading hf_secret_token from the truss secret and setting it with os.environ["HF_TOKEN"] we don't need to store the HF token as plaintext in config and can use secrets

dsingal0:
> Do you think this pattern is expected? It looks like the instruct response completes part of the request (! 😀), which makes me think there is potentially a templating issue?
> 
> image

I believe this is because the vllm implementation didn't use the chat template. I'm not familiar with that, but will give it a try.

gemma2/gemma2-27b-it-vllm/model/model.py
vshulman: hmmm, what would logging in to the HF hub accomplish here? I think I was able to deploy this successfully by commenting out L25

vshulman (approved):

gemma2/gemma2-27b-it-vllm/model/model.py
dsingal0: Might've been an authentication fluke. COnfirmed it still works without login, and also works with autotokenizer for gemma2

PR #324, opened 2024-07-16
reorganized gemma2 under gemma

dsingal0:
Reorganized gemma 2 under gemma models. Added repo_id to metadata to have correct terms and conditions on the model access under model library.


gemma/gemma-2-27b-it-vllm/config.yaml
```diff
@@ -3,6 +3,7 @@ python_version: py311
 model_metadata:
   example_model_input: {"prompt": "what is the meaning of life"}
   main_model: google/gemma-2-27b-it
+  repo_id: https://huggingface.co/google/gemma-2-27b-it
```
vshulman, line 6: by convention we only do `google/gemma-2-27b-it`; i wonder if we can get rid of main_model above, and just reference `repo_id`?

gemma/gemma-2-9b-it-vllm/config.yaml
```diff
@@ -3,6 +3,7 @@ python_version: py311
 model_metadata:
   example_model_input: {"prompt": "what is the meaning of life"}
   main_model: google/gemma-2-9b-it
+  repo_id: https://huggingface.co/google/gemma-2-9b-it
   tensor_parallel: 1
```
vshulman, line 7: same note as above

vshulman (changes requested):

vshulman (approved):

PR #327, opened 2024-07-23
added llama 3.1 8b instruct vllm 

dsingal0:


llama/llama-3_1-8b-instruct/config.yaml
```diff
@@ -0,0 +1,21 @@
+model_name: "Llama 3.1 8B Instruct VLLM"
+python_version: py311
+model_metadata:
+  example_model_input: {"prompt": "what is the meaning of life"}
+  repo_id: meta-llama/Meta-Llama-3.1-8B-Instruct
+  tensor_parallel: 1
+requirements:
+  - vllm==0.5.3
+  - https://github.com/flashinfer-ai/flashinfer/releases/download/v0.0.9/flashinfer-0.0.9+cu121torch2.3-cp311-cp311-linux_x86_64.whl
```
vshulman, line 9: can be removed?

vshulman (approved):
lgtm -- i think we don't need flash infer anymore

PR #335, opened 2024-08-07
generic vllm example

tianshuc0731:
**Motivation**
Create this generic vllm example to enable a more codeless customer experience to use truss with vllm. The goal is to let customers create any customized vllm deployment by only modifying the config file, no need to touch model.py.

**Changes**
In `config.yaml`
* add new field `openai_compatible` to specify if users want to use [openai compatible mode](https://docs.vllm.ai/en/stable/serving/openai_compatible_server.html) with vllm
* add new field `vllm_config` to take all [vllm engine parameter](https://docs.vllm.ai/en/stable/models/engine_args.html) override for users

In `model.py`
Combine two modes of using vllm in one file
* when `openai_compatible = false` (standard mode), load() will initiate `AsyncLLMEngine`
* when `openai_compatible = true` (openai compatible mode), load() will skip initiating `AsyncLLMEngine` but start a openai compatible inference server locally

**Testing**
Successfully used this example to deploy the following models:
* llama3.1 8B Instruct openai compatible mode 
* llama3.1 8B Instruct standard mode 
* Gemma 2 9B Instruct standard mode 
* Mistral 7B v2 vLLM AWQ - T4 (model quantization)
* Ultravox v0.2 (customized vllm image)

**To follow up**
* Figure out better way for server startup failure than using `MAX_FAILED_SECONDS`
* Add speculative decoding example to README
* Add health check to monitor vllm openai server


11-using-vllm/model/model.py
```diff
@@ -0,0 +1,182 @@
...
+                    json=model_input,
+                )
+                return response.json()
+        else:
+            prompt = model_input.pop("prompt")
+            stream = model_input.pop("stream", True)
+
+            sampling_params = SamplingParams(**model_input)
+            idx = str(uuid.uuid4().hex)
+            chat = [
```
vshulman, line 157: I think folks are increasingly expecting a messages input as a valid input in addition to prompt, even though originally it was meant for OpenAI. What do you think about adding support for it, similar to what we do [here](https://github.com/basetenlabs/truss-examples/blob/main/lora/model/model.py).

vshulman (changes requested):
Overall I think this is solid. Let's definitely add a README to describe all the different features we have here. You can base it off another README that you find in truss-examples, the ultravox one might work. I might also suggest including a few example configs in the README for common settings, maybe prefix caching and quantization

The readme should also talk about how you can use the bridge + openai compatible mode to use the truss with openAI lib

I think the numbered repos (e.g. 10-using-system-packages) correspond to examples in the docs (e.g. https://docs.baseten.co/deploy/examples/10-using-system-packages). What do you think about removing [this](https://github.com/basetenlabs/truss-examples/tree/main/vllm-quantization/model) example, and adding yours simply as "vllm"?

11-using-vllm/model/model.py
```diff
@@ -0,0 +1,182 @@
...
+            )
+
+            process = subprocess.Popen(
+                command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True
+            )
+
+            # Wait for 10 seconds and check if command fails
+            time.sleep(10)
+
+            if process.poll() is None:
```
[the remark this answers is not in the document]
tianshuc0731: we actually only use it once when we use subprocess to start openai compatible server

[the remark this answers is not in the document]
tianshuc0731: punt for now

11-using-vllm/model/model.py
```diff
@@ -0,0 +1,182 @@
...
+from vllm.engine.arg_utils import AsyncEngineArgs
+from vllm.engine.async_llm_engine import AsyncLLMEngine
+
+os.environ["TOKENIZERS_PARALLELISM"] = "true"
+
+logger = logging.getLogger(__name__)
+
+
+class Model:
+    MAX_FAILED_SECONDS = 600  # 10 minutes; the reason this would take this long is mostly if we download a large model
```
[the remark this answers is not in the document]
tianshuc0731: I tested it with 70B and it worked, punt the investigation into better practice for now due to bandwidth limit, but will create a ticket to follow up.

11-using-vllm/model/model.py
```diff
@@ -0,0 +1,182 @@
...
+                    "Server failed to start within the maximum allowed time."
+                )
+        else:
+            try:
+                result = subprocess.run(
+                    ["nvidia-smi"], capture_output=True, text=True, check=True
+                )
+                print(result.stdout)
+            except subprocess.CalledProcessError as e:
+                print(f"Command failed with code {e.returncode}: {e.stderr}")
```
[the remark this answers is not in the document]
tianshuc0731: done

11-using-vllm/model/model.py
```diff
@@ -0,0 +1,182 @@
...
+
+    async def predict(self, model_input):
+        if self.openai_compatible:
+            # if the key metrics: true is present, let's return the vLLM /metrics endpoint
+            if model_input.get("metrics", False):
+                response = await self._client.get(f"{self.vllm_base_url}/metrics")
+                return response.text
```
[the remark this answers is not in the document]
tianshuc0731: done

11-using-vllm/model/model.py
tianshuc0731: done

vllm/model/model.py
```diff
@@ -0,0 +1,204 @@
...
+            logger.info(
+                f"Starting openai compatible vLLM server with command: {command}"
+            )
+
+            process = subprocess.Popen(
+                command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True
+            )
+
+            # Wait for 10 seconds and check if command fails
+            time.sleep(10)
```
vshulman, line 64: I might need some help understanding this flow

vllm/model/model.py
```diff
@@ -0,0 +1,204 @@
...
+            self.vllm_base_url = f"http://localhost:{self._vllm_port}"
+
+            # Polling to check if the server is up
+            server_up = False
+            start_time = time.time()
+            while time.time() - start_time < self.MAX_FAILED_SECONDS:
+                try:
+                    response = httpx.get(f"{self.vllm_base_url}/health")
+                    logger.info(
+                        f"Checking server health: {response.status_code}, {response.text}"
```
vshulman, line 90: nit: might be overly verbose?

vllm/model/model.py
```diff
@@ -0,0 +1,204 @@
...
+                return response.json()
+        else:
+            # SamplingParams does not take/use argument 'model'
+            if "model" in model_input:
+                model_input.pop("model")
+            if "prompt" in model_input:
+                prompt = model_input.pop("prompt")
+                sampling_params = SamplingParams(**model_input)
+                idx = str(uuid.uuid4().hex)
+                messages = [
```
vshulman, line 170: do you feel strongly about this? I can also see an argument for setting input = prompt directly, i.e. if the user wants to deal with templating on their side

vllm/README.md
```diff
@@ -0,0 +1,206 @@
+# vLLM Truss
+
+## What is this Truss example doing
+
+This is a generic [Truss](https://truss.baseten.co/) that can deploy an asynchronous vLLM engine([AsyncLLMEngine](https://docs.vllm.ai/en/latest/dev/engine/async_llm_engine.html#asyncllmengine)) of any customized configuration with [all compatible models](https://docs.vllm.ai/en/latest/models/supported_models.html). We create this example to give you the most codeless experience, so you can configure all vLLM engine parameters in `config.yaml`, without making code changes in `model.py` for most of the use cases.
```
vshulman, line 5: maybe "general purpose" instead of generic?

vllm/README.md
```diff
@@ -0,0 +1,206 @@
...
+  use_gpu: true
+runtime:
+  predict_concurrency: 128
+secrets:
+  hf_access_token: null
+```
+
+### Use vLLM's OpenAI compatible server
+
+To use vLLM in [OpenAI compatible server](https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html) mode, simply set `openai_compatible: true` under `model_metadata`.
```
vshulman, line 64: let's include a link to our openai bridge docs

vllm/README.md
```diff
@@ -0,0 +1,206 @@
...
+model_name: "Llama 3.1 8B Instruct VLLM"
+python_version: py311
+model_metadata:
+  example_model_input: {"prompt": "what is the meaning of life"}
+  repo_id: meta-llama/Meta-Llama-3.1-8B-Instruct
+  openai_compatible: false
+  vllm_config:
+    tensor_parallel_size: 4
+    max_model_len: 4096
+    distributed_executor_backend: mp
```
vshulman, line 50: what do you think about optionally defaulting to some backend in model.py if this isn't configured explicitly?

vshulman (commented):
Looks good, and thanks for updating the other examples
I left a few suggestions + questions

vllm/model/model.py
```diff
@@ -0,0 +1,204 @@
...
+import uuid
+
+import httpx
+from transformers import AutoTokenizer
+from vllm.engine.arg_utils import AsyncEngineArgs
+from vllm.engine.async_llm_engine import AsyncLLMEngine
+
+from vllm import SamplingParams
+
+os.environ["TOKENIZERS_PARALLELISM"] = "true"
```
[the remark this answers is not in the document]
tianshuc0731: do any of you @htrivedi99 @dsingal0 know if this is necessary?

vllm/model/model.py
```diff
@@ -0,0 +1,204 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Model:
+    # TODO: better way to detect model server startup failure than using `MAX_FAILED_SECONDS`
+    MAX_FAILED_SECONDS = 600  # 10 minutes; the reason this would take this long is mostly if we download a large model
```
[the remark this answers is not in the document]
tianshuc0731: I tested 70B model takes ~3 mins
Extend to 25 mins to accommodate model size up to 405B

vllm/model/model.py
tianshuc0731: Wait for 10 seconds and then check if `vllm` command is still running, to early detect any command running failure

vllm/README.md
tianshuc0731: by default vllm uses "ray" if installed otherwise "mp"

vllm/model/model.py
tianshuc0731: I would prefer not to do that cause I want to make API as simple as possible if it can cover 90% use case. Supporting raw input may need to add an extra flag in the request body like "enable_raw_input".
If we see high customer demand for this use case, I will change my mind.

vshulman (approved):

PR #337, opened 2024-08-19
add max tokens to llama 3.1 example

philipkiely-baseten:


relph (approved):

dsingal0:
@philipkiely-baseten can you add it to 70B and 405B as well?

PR #338, opened 2024-08-20
vllm health check

tianshuc0731:
**Motivation**
To enable truss auto detect vllm engine crash down and self recover through restart.

**Changes**
At the end of model loading step, we start a long-running background thread to probe vllm engine/server health every  `HEALTH_CHECK_INTERVAL` seconds, and trigger `os.exit(1)` upon health check failure.

**Testing**

For standard and openai server mode, truss can auto detect vllm engine failure and restart successfully.

Use case 1: standard vllm engine mode, fake engine crash by shut down the vllm engine event loop

Screenshot 2024-08-19 at 7 35 50 PM

Use case 2: openai compatible server mode, fake server crash by kill the server process

Screenshot 2024-08-19 at 6 39 20 PM


vllm/model/model.py
```diff
@@ -32,6 +35,32 @@ def __init__(self, **kwargs):
...
+            )
+            os._exit(1)
+
+    def _run_background_vllm_health_check(self):
+        logger.info("Starting background health check loop")
+        loop = asyncio.new_event_loop()
+        loop.create_task(self._monitor_vllm_health())
+        thread = threading.Thread(target=loop.run_forever, daemon=True)
+        thread.start()
+
```
htrivedi99, line 63: Do you think it makes sense to move these functions over to another file? It would make our model.py look a lot cleaner

tianshuc0731: I am open to that.

htrivedi99 (approved):
lgtm

PR #347, opened 2024-09-11
[BT-12061] Add phi 3.5 mini instruct example

derrick-yyang:
Used vLLM template to add Phi 3.5 mini-instruct model

Note: ignore the commit for using port `8080`. The default port `8000` is used


phi/phi-3.5-mini/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+### OpenAI SDK (if you are using OpenAI compatible server)
+
+```
+from openai import OpenAI
+import os
+
+model_id = "a2345678" # Replace with your model ID
+
+client = OpenAI(
```
vshulman, line 69: I think this OpenAI syntax actually changed slightly in recent weeks:
https://docs.baseten.co/api-reference/openai

phi/phi-3.5-mini/config.yaml
```diff
@@ -0,0 +1,18 @@
...
+  example_model_input: {"prompt": "what is the meaning of life"}
+  repo_id: microsoft/Phi-3.5-mini-instruct
+  openai_compatible: true
+  vllm_config:
+    tensor_parallel_size: 1
+    max_model_len: 10000
+requirements:
+  - vllm==0.5.4
+resources:
+  accelerator: A100
```
vshulman, line 13: let's use a smaller GPU, since the beauty of a smaller model is that we can do that!

vshulman (changes requested):
can we think of a way to reuse the `model.py` file without directly copying it?

phi/phi-3.5-mini/README.md
derrick-yyang: do you know the exact change? I don't see any difference between the api docs you sent me and the code in the readme

vshulman: at a minimum, the `extra_body` field
might be worth testing this yourself to make sure this works

derrick-yyang: updated both readmes 👍

vshulman (approved):
lgtm!

PR #348, opened 2024-09-12
Fixed stream=True option for VLLM template

derrick-yyang:
Currently when we process `model_input`, we popped the "stream" property. 

We want to maintain this property when openAI compatible mode is enabled so stream works with openAI SDK calls


htrivedi99 (approved):
Looks good!

PR #354, opened 2024-09-24
Cancel pipeline. Doesn't work for request canlleations yet

marius-baseten:


flux/schnell/model/model.py
```diff
@@ -41,11 +47,35 @@ def convert_to_b64(self, image: Image) -> str:
...
+                generator=generator,
+            ).images[0]
+            self._result = image
+            return
+        except Exception as e:
+            logging.info(f"Image generation was aborted or failed: {e}")
+            self._exception = e
+            return
+
+    async def predict(self, model_input, request: fastapi.Request):
```
squidarth, line 72: I wonder if we can keep this sync, and use something like this to check for disconnection. ie:

```
from anyio import from_thread
def predict(self, model_input, request: fastapi.Request):
     is_disconnected = from_thread.run(request.is_disconnected)
     if is_disconnected:
         raise HTTPException(...)    
```

cancelling the ongoing inference is tricky though

PR #368, opened 2024-10-18
Added Qwen 2.5 7B Instruct

dsarfati:
This PR adds Qwen 2.5 7B Instruct using trussless vllm. This model is planned to be added to the model library

### DIFF
--- a/qwen/qwen-2_5-7b-instruct-vllm/config.yaml
+++ b/qwen/qwen-2_5-7b-instruct-vllm/config.yaml
@@ -1,0 +1,33 @@
+environment_variables: {}
+external_package_dirs: []
+model_metadata:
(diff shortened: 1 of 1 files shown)


qwen/qwen-2_5-7b-instruct-vllm/config.yaml
```diff
@@ -0,0 +1,33 @@
+environment_variables: {}
+external_package_dirs: []
+model_metadata:
+  avatar_url: https://cdn.baseten.co/production/static/explore/meta.png
```
vshulman, line 4: these are probably wrong :)

vshulman (changes requested):

philipkiely-baseten:
Added via engine buider

PR #373, opened 2024-11-01
Improve truss examples CI script

squidarth:
Improves the truss examples to do the following:
* Fail immediately if there are problems (instead of having to wait 20 minutes)
* Doesn't have a 20 minute timeout if the model is still building

# Testing

Ran https://github.com/basetenlabs/truss-examples/actions/runs/11633933987/job/32400112601

### DIFF
--- a/bin/test_example.py
+++ b/bin/test_example.py
@@ -3,6 +3,7 @@ import sys
 import time
 
 import requests
+import truss
 from tenacity import (
     Retrying,
     retry,
@@ -30,7 +31,7 @@ def write_trussrc_file(api_key: str):
     RemoteFactory.update_remote_config(ci_user)
 
 
-@retry(wait=wait_fixed(60), stop=stop_after_attempt(20), reraise=True)
+@retry(wait=wait_fixed(30), stop=stop_after_attempt(3), reraise=True)
 def attempt_inference(truss_handle, model_version_id, api_key):
     """
     Retry every 20 seconds to call inference on the example, using the `example_model_input`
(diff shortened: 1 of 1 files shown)


bin/test_example.py
```diff
@@ -65,26 +66,24 @@ def attempt_inference(truss_handle, model_version_id, api_key):
...
+@retry(
+    wait=wait_random_exponential(multiplier=1, max=120),
+    stop=stop_after_attempt(3),
+    reraise=True,
+)
+def deploy_truss(target_directory: str) -> str:
+    model_deployment = truss.push(
+        target_directory, remote=REMOTE_NAME, trusted=True, publish=True
+    )
+    model_deployment.wait_for_active()
```
bdubayah, line 78: Nice, so this will block until the model comes up?

bdubayah (approved):
Awesome, LGTM!

bin/test_example.py
squidarth: yep, and fail fast  as soon as there's a failure

marius-baseten: Could you import this from it's defining module instead? I.e. `from truss.api...`? Forward-compatible with the separation of concerns in the truss package.

marius-baseten:
Great quality of life improvement, thanks!

marius-baseten (approved):

bin/test_example.py
squidarth: Wait -- I thought from your PR we were leaving truss.push in the top-level? https://github.com/basetenlabs/truss/pull/1206/files#diff-92d69f0f5363f1d4e214eaf4a10c951888cf1a388ee43f4ed7cd57e720e837e1

There are several users using `truss.push` in their CI jobs now, so removing this would be a pretty serious breaking change (we've documented this API) https://docs.baseten.co/truss-reference/python-sdk

squidarth: I'm going to merge this for now since something seems quite broken with the current script (which calls the legacy predict endpoints), but let's discuss this further, want to make sure we're aligned (and can change if we have to)

PR #379, opened 2024-11-18
Update Llama 3.2 11B Vision Instruct to run with `vllm serve`

parallelipopat:
Credit to @htrivedi99.


llama/llama-3_2-11b-vision-instruct/config.yaml
```diff
@@ -1,7 +1,8 @@
-model_name: "Llama 3.2 11B Vision Instruct VLLM openai compatible"
-python_version: py311
+base_image:
+  image: vllm/vllm-openai:v0.6.3.post1
 model_metadata:
   example_model_input: {
+    model: "llama-3.2-11b-vision-instruct",
     messages: [
       {
         role: "user",
```
htrivedi99, line 8: I think it might be better to have a prompt analyzing an image since this is a vision language model.

parallelipopat: The prompt here is to describe an image
```
{
            type: "text",
            text: "Describe this image in one sentence."
},
```

Was this comment for Line 40 of README.md that calls `truss predict`? There's an example of calling the model  to describe an image below that line; I can just delete the `truss predict` option.

htrivedi99: Oh wait, nevermind. I didn't see the whole config.
You're good

htrivedi99 (approved):

PR #386, opened 2024-12-06
text-embeddings-inference updated example trussless

michaelfeil:
adding a trusses example for text-embeddings-inference


text-embeddings-inference/custom_server/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+1. Make sure you have a [Baseten account](https://app.baseten.co/signup) and [API key](https://app.baseten.co/settings/account/api_keys).
+2. Install the latest version of Truss: `pip install --upgrade truss`
+3. [Required for gated model] Retrieve your Hugging Face token from the [settings](https://huggingface.co/settings/tokens). Set your Hugging Face token as a Baseten secret [here](https://app.baseten.co/settings/secrets) with the key `hf_access_key`.
+
+First, clone this repository:
+
+```sh
+git clone https://github.com/basetenlabs/truss-examples.git
+cd custom-server/text-embeddings-inference/custom_server
```
vshulman, line 17: is this actually nested within customer-server?

text-embeddings-inference/custom_server/README.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+With `custom_server` as your working directory, you can deploy the model with the following command, paste your Baseten API key if prompted.
+
+```sh
+truss push --publish --trusted
```
vshulman, line 23: can remove trusted now

text-embeddings-inference/custom_server/.maintain/roll_out_docker.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
vshulman, line 1: do we want to keep this?

text-embeddings-inference/custom_server/.maintain/roll_out_docker.sh
```diff
@@ -0,0 +1,28 @@
...
+  ["turing"]="turing-"
+  ["ampere80"]=""
+  ["ampere86"]="86-"
+  ["adalovelace"]="89-"
+  ["hopper"]="hopper-"
+)
+
+# Define version and target
+VERSION="1.5"
+TARGET="michaelf34/text-embeddings-inference"
```
vshulman, line 16: nit: should we move this into the baseten docker namespace?

text-embeddings-inference/custom_server/README.md
```diff
@@ -0,0 +1,80 @@
...
+        -H "Authorization: Api-Key YOUR_API_KEY" \
+        -d '{"input": "text string"}'
+```
+
+### request python library
+
+```python
+import requests
+
+resp = requests.post(
```
vshulman, line 41: it would be helpful to make this a more specific example for embeddings, e.g. passing in multiple values at the same time

text-embeddings-inference/custom_server/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+client = OpenAI(
+    api_key=os.environ["YOUR_API_KEY"],
+    base_url="https://bridge.baseten.co/v1/direct"
+)
+
+model_id = "xxx"
+deployment_id = "xxx"
+
+response = client.embeddings.create(
```
vshulman, line 64: as discussed: great if this works out of the box with openai python library; if not, we should remove or clarify how to get that behavior

text-embeddings-inference/custom_server/config.yaml
```diff
@@ -0,0 +1,29 @@
...
+  # Turing (T4, ...)	michaelf34/text-embeddings-inference:turing-1.5
+  # Ampere 80 (A100, A30)	michaelf34/text-embeddings-inference:1.5
+  # Ampere 86 (A10, A10G, A40, ...)	michaelf34/text-embeddings-inference:86-1.5
+  # Ada Lovelace (L4, ...)	michaelf34/text-embeddings-inference:89-1.5
+  # Hopper (H100/H100 40GB)	michaelf34/text-embeddings-inference:hopper-1.5
+  image: michaelf34/text-embeddings-inference:1.5
+model_metadata:
+  repo_id: BAAI/bge-base-en-v1.5
+docker_server:
+  start_command: sh -c "text-embeddings-router --port 7998 --model-id /data/local-model --max-client-batch-size 256 --max-concurrent-requests 40 --max-batch-tokens 32768"
```
vshulman, line 13: any guidance on these numbers? Maybe something we can call out in the README for how to set them

vshulman (changes requested):
A few nits -- I think the biggest one is we should make the README a little more embeddings-specific

text-embeddings-inference/custom_server/README.md
michaelfeil: Makes sense!

text-embeddings-inference/custom_server/README.md
michaelfeil: Fixed it

text-embeddings-inference/custom_server/.maintain/roll_out_docker.sh
michaelfeil: Not sure, that's why I left it in. Do we want to keep this?

text-embeddings-inference/custom_server/README.md
michaelfeil: I tried out some options & none of them worked. So I removed it.

text-embeddings-inference/custom_server/config.yaml
michaelfeil: Wrote a short essay in the readme!

michaelfeil:
Okay, worked on it!

internal/config.yaml
vshulman, line 1: stray file?

text-embeddings-inference/README.md
```diff
@@ -1,32 +1,109 @@
-# Text Embeddings Inference Truss (A100)
-This is an example of a Truss model that uses the Text Embeddings Inference API.
+# Text Embeddings Inference Truss
 
-## How to Deploy
-In the `config.yaml` file, you can specify the model to use, as well as other arguments per the [Text Embeddings Inference API](https://huggingface.co/docs/text-embeddings-inference) documentation.
-Note that not all models are supported by TEI.
+This is a [Truss](https://truss.baseten.co/) to deploy [text-embeddings-inference](https://github.com/huggingface/text-embeddings-inference), a high performance embedding and reranking api.
```
vshulman, line 3: maybe we should be saying "this is a Custom Server to deploy..."

text-embeddings-inference/README.md
```diff
@@ -1,32 +1,109 @@
...
+```
+
+## Performance Optimization:
+
+The config.yaml contains a couple of variables that can be tuned, depending on:
+- which GPU is used
+- which model is deployed
+- how many concurrent requests users are sending
+
+The deployment example is for Bert-large and a Nvidia-L4. Bert-large has a maxiumum sequence length of 512 tokens per sentence.
```
vshulman, line 33: Bert=>BERT

text-embeddings-inference/README.md
```diff
@@ -1,32 +1,109 @@
...
 ```
 
-## How to Generate Embeddings
-The truss expects:
-- "texts" parameter with either a single string or an array of strings.
-- "stream" parameter with a boolean value (default is false).
+This number of total tokens in a batch. For embedding models, this will determine the VRAM usage.
+As most of TEI's models are implemented with `nested` attention implementation, `32768 tokens` could mean `64 sentence with 512 tokens` or `512 sentences with 64 tokens`. While the first will take slightly longer to compute, the peak VRAM usage will stay roughly the same. For `llama` or `mistral` based `7b` embedding models, we recommend setting it a lower setting e.g.
+```
+--max-batch-tokens 8192
```
vshulman, line 55: very helpful

text-embeddings-inference/README.md
```diff
@@ -1,32 +1,109 @@
...
+```
+
+### Client batch size
+```
+--max-client-batch-size 256
+```
+This determines the number of sentences / items in a single request.
+For optimal autoscaling that gets regulated by metrics such as requests/second in Baseten's infrastructure, you want to set this as low as possible. OpenAI-API historically set it to `--max-client-batch-size 32`, which could help for more aggressive autoscaling and thus better latency. One the other hand, frameworks such as LLamaIndex, Langchain or Haystack might prefer or even require higher batch_sizes, especially if the user code is old-fashioned and sends requests 1-by-1 in a for loop. This depends on your users & how you are planning to use your deployment.
```
vshulman, line 63: this is a little hard to follow

vshulman (commented):
One stray file (I think) and two copy changes requested. Rest LGTM.

text-embeddings-inference/README.md
michaelfeil: updated it, and just set it 32.

PR #391, opened 2024-12-18
feat: support deepseek v2.5 1210 with sglang

zhyncs:


zhyncs:
BF16 will run out of memory, so I'm trying to use `llmcompressor` to get an FP8 model.

michaelfeil (approved):

PR #402, opened 2025-01-23
Update pyproject.toml + tools Falcon 3 addition

michaelfeil:
- tools are super outdated
- poetry pins `truss==0.9.49` via git! Git still pins revisions!

Adds falcon:
- falcon 3 in favor of falcon-40B


philipkiely-baseten (approved):
Looks good. Which config would you recommend for the library?

falcon/falcon3-10B-trt-llm-spec-dec/config.yaml
```diff
@@ -0,0 +1,48 @@
...
+  cpu: "1"
+  memory: 24Gi
+  use_gpu: true
+trt_llm:
+  build:
+    base_model: llama
+    checkpoint_repository:
+      repo: tiiuae/Falcon3-10B-Instruct
+      source: HF
+    max_seq_len: 8192
```
philipkiely-baseten, line 29: Should this match the runtime tokens?

falcon/falcon3-10B-trt-llm-spec-dec/config.yaml
```diff
@@ -0,0 +1,48 @@
...
+        # This is the checkpoint repository for the speculative decoding model.
+        # Models of the same architecture may be fine-tuned for higher performance
+        repo: tiiuae/Falcon3-1B-Instruct
+        source: HF
+      num_draft_tokens: 4
+  runtime:
+    enable_chunked_context: true
+    kv_cache_free_gpu_mem_fraction: 0.62
+    request_default_max_tokens: 1000
+    total_token_limit: 500000
```
philipkiely-baseten, line 48: Should this be a round number?

falcon/falcon3-10B-trt-llm-spec-dec/config.yaml
michaelfeil: This should match: https://huggingface.co/tiiuae/Falcon3-1B-Instruct/blob/28ba2251970a01dd1edc7ba7dad2eb71216ccfdf/config.json#L13

falcon/falcon3-10B-trt-llm-spec-dec/config.yaml
michaelfeil: total_token_limit defines a briton setting for how many tokens are queued inside C++. If queueing too many requests, we overload runtime. Not important, as long as it's in a decent range. 

total_token_limit = 500000 is default.

kv_cache_free_gpu_mem_fraction: 0.62 is a hard setting. 
Falcon-3-10B needs around 20GB Vram for weights. We have 20GB Vram left. 
Of the rest of the 20GB we allocate 62% to Falcon-10B. The rest of 7.6GB go to the 1B model. 2 GB out of the 7.6Gb will go to the weights, while 5.6GB are for activations and KV cache.

PR #408, opened 2025-02-13
Remove bark

squidarth:
This example is no longer working. It is not referenced in the docs anywhere.


philipkiely-baseten (approved):
I'm good with this -- planning a bigger cleanup of these examples soon.

PR #414, opened 2025-02-18
[Truss Examples CI] delete all but last 5 versions in each test case.

marius-baseten:


squidarth (approved):

bin/test_example.py
```diff
@@ -76,15 +72,45 @@ def deploy_truss(target_directory: str) -> str:
...
+
+def clean_up_deployments(model_name: str, model_id: str):
+    remote: BasetenRemote = RemoteFactory.create(REMOTE_NAME)
+    model = remote.api.get_model(model_name)
+    versions = [
+        (version["id"], version["semver"])
+        for version in model["model"]["versions"]
+        if not version["is_primary"]
+    ]
+    versions = sorted(versions, key=lambda x: x[1])
```
moward, line 94: Is lexicographically sorting by the semver always going to give you the right order? For example, "0.10" < "0.2", but in semver it should be the other way.

marius-baseten: Good point. Parsed the version.

bin/test_example.py
```diff
@@ -76,15 +73,45 @@ def deploy_truss(target_directory: str) -> str:
...
 
+def delete_model_deployment(api: BasetenApi, model_id: str, deployment_id: str) -> Any:
+    url = f"{api._rest_api_url}/v1/models/{model_id}/deployments/{deployment_id}"
+    headers = api._auth_token.header()
+    resp = requests.delete(url, headers=headers)
+    if not resp.ok:
+        resp.raise_for_status()
+
+    deployment = resp.json()
+    return deployment
```
nnarayen, line 84: nit: Do we need any return here? Seems like we don't consume it below

nnarayen (approved):
Awesome!

PR #416, opened 2025-03-12
added gemma 3

dsingal0:


gemma/gemma-3-27b-it/config.yaml
```diff
@@ -0,0 +1,54 @@
...
+        {
+        "type": "image_url",
+        "image_url": {
+          "url": "https://picsum.photos/id/237/200/300"
+        }
+        }
+      ]
+      }
+    ],
+    "stream": false,
```
philipkiely-baseten, line 26: Why false by default?

philipkiely-baseten (approved):
Thank you!

gemma/gemma-3-27b-it/config.yaml
dsingal0: no reason, will make it true

PR #420, opened 2025-03-17
mistral 3.1 small

dsingal0:
for real use cases we'll need to raise to TP2 to enable higher sequence length and more images per request, but in-line with gemma 27B, pushing with TP 1 H100.


philipkiely-baseten (approved):
LGTM, deploy in Billip please?

PR #429, opened 2025-04-24
fix streaming issue for token gen

veerbia:
Correct `token_gen` handling to grab body_iterator in the case we have that it's an instance of `StreamingResponse`.


michaelfeil (commented):
I'll fix this in a follow up PR.

PR #435, opened 2025-04-28
Fix lint, let it run on a later version of ubuntu

squidarth:
Make sure the lint can run. This was running on an old version of ubuntu.


marius-baseten (approved):

marius-baseten:
Should we update this everywhere?

PR #439, opened 2025-05-08
add no stuttering orpheus

michaelfeil:


orpheus-best-performance/call.py
```diff
@@ -7,11 +7,11 @@
 
 # Configuration
 MODEL = "dq4rlnkw"
-BASETEN_HOST = f"https://model-{MODEL}.api.baseten.co/environments/production/predict"
+BASETEN_HOST = "https://model-dq4rlnkw.api.baseten.co/deployment/wl6pl1w/predict"
```
veerbia, line 10: maybe need to revert this to avoid hardcoding

veerbia (approved):

PR #448, opened 2025-05-21
Ultravox 0.4 fixes and audio support

ShivainVij:
Removed HF Access Token (not needed, just causes problems for the customer)

Changed example audio to be a public URL (baseten cdn at this time has a robots.txt)

Added audio support in requirements.


parallelipopat (approved):
Nice fix!

PR #474, opened 2025-07-28
Demonstrate how to deploy a model on trtllm with multiple loras

FredLiu876:
This is an example of a truss that supports dynamic swapping of LoRA adapters to serve multiple fine tuned models from a single GPU. This example deploys a Mistral 7B Instruct that is an expert in finance, medicine and law.


dsarfati (approved):

FredLiu876:
@JimmyWhitaker I think that's an excellent point, and if we have more demand the company should definitely build some custom solutions for LoRA. Let's keep note of that.

PR #475, opened 2025-07-29
Update `--index-url` examples, with move to `uv`

nnarayen:
[Move](https://basetenlabs.slack.com/archives/C04NM3YA2DT/p1753795279056229) from `pip` to `uv` as our base installer introduced a couple issues with usage of `--index-url`. While we track down who might have run into these issues, we should update our examples so no new users have issues.


squidarth (approved):

michaelfeil:
WAIT, this fix will not work!

PR #480, opened 2025-07-31
GLM 4.5 examples

philipkiely-baseten:
Add examples for Air and full GLM 4.5. Will be adding Air to the model library, full is just there as a reference as it has major hardware requirements.


michaelfeil (approved):
go ahead!

PR #481, opened 2025-07-31
Added llama models and standardized format.

JimmyWhitaker:


FredLiu876:
Thanks for renaming the folder and having that folder level README! Very nice

PR #581, opened 2026-07-08
Add inference SSH tune-vllm-args example

cretz:
## What changed

Adds a new `inference-ssh/` example category demonstrating an in-container
development loop over SSH, plus its first example, `tune-vllm-args`.

- Deploys a small vLLM server (Qwen2.5-0.5B on an L4) as a custom `docker_server` with `runtime.remote_ssh.enabled`.
- The server runs under `entr`, which watches `data/` and restarts vLLM in place whenever a file changes, so edits over SSH take effect without redeploying.
- Walkthrough README covers SSH setup with Baseten CLI, deploy, pinning a replica, calling the model, and the SSH edit/restart loop, using a chat-template persona swap (pirate to Shakespeare) as the visible example.
- Adds an `inference-ssh/` index README linking the SSH docs.

Note, an earlier version of this worked with `truss` CLI in addition to `baseten` CLI, but now the `truss` one has been removed

### DIFF
--- a/inference-ssh/tune-vllm-args/README.md
+++ b/inference-ssh/tune-vllm-args/README.md
@@ -17,46 +17,18 @@ on a single L4 GPU.
 
 ## 1. Set up SSH
 
-Set up SSH once per machine, using the
-[Truss CLI](https://docs.baseten.co/reference/cli/truss/overview) or the
-[Baseten CLI](https://docs.baseten.co/reference/cli/baseten/overview). Install
-whichever you prefer by following its docs, then run its setup below. For background
+Set up SSH once per machine with the
+[Baseten CLI](https://docs.baseten.co/reference/cli/baseten/overview). For background
 on SSH access, see the [SSH docs](https://docs.baseten.co/inference/ssh).
-
-**Using the Truss CLI**
-
-```sh
-truss login
-truss ssh setup
-```
-
-When `truss login` prompts for an authentication method, choose the API key option.
-SSH signing needs the API key stored locally, so browser login alone will not work
-for `truss ssh`.
-
-**Using the Baseten CLI**
 
 ```sh
 baseten auth login
 baseten ssh setup
 ```
 
-> **Note:** Baseten CLI SSH is not released yet, so the Baseten CLI must be
-> [built from source](https://github.com/basetenlabs/baseten-cli#building) and put on
-> PATH to use it.
-
 ## 2. Deploy
 
-Push with the Truss CLI or the Baseten CLI. Both accept `--wait` to block until the
-deployment is `ACTIVE`.
-
-**Using the Truss CLI**
-
-```sh
-truss push inference-ssh/tune-vllm-args --wait
-```
-
-**Using the Baseten CLI**
+Push with the Baseten CLI. Pass `--wait` to block until the deployment is `ACTIVE`.
 
 ```sh
 baseten model push --dir inference-ssh/tune-vllm-args --wait
@@ -85,7 +57,7 @@ export MY_MODEL_ID=<model_id>
 
 **Using curl**
 
-Set your Baseten API key too.
+Using curl needs an API key:
 
 ```sh
 export MY_API_KEY=<your-api-key>
--- a/inference-ssh/tune-vllm-args/README.md
+++ b/inference-ssh/tune-vllm-args/README.md
@@ -1,0 +1,182 @@
+# Tune vLLM args over SSH
+
+This example demonstrates an in-container development loop over SSH. It deploys a small
+vLLM server with SSH access enabled, then shows how to SSH into a running replica, edit
+files in place, and restart the app to pick up the changes without redeploying or
+recreating the container.
+
+The server is launched by [`entr`](https://eradman.com/entrproject/), which watches
+the files in `/app/data` and relaunches vLLM whenever one of them changes. Every
+vLLM flag lives in [`data/vllm_args.txt`](data/vllm_args.txt), and the assistant
+persona lives in [`data/chat_template.jinja`](data/chat_template.jinja). See
+[`config.yaml`](config.yaml) for how it is wired together. SSH into the replica, edit
+either file, and `entr` restarts the vLLM engine in place.
+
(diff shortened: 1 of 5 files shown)


inference-ssh/tune-vllm-args/README.md
cretz, line 1: This file will read much simpler when `baseten` is the stable CLI of choice

inference-ssh/tune-vllm-args/data/chat_template.jinja
marius-baseten, line 1: Isn't the persona/system prompt normally inserted via reqeust field instead of baked into a static template? Side node: jinka is per se a total pain IMO, so moving anything out there that doesn't have to be there for other reasons is a win.

inference-ssh/tune-vllm-args/README.md
```diff
@@ -0,0 +1,182 @@
+# Tune vLLM args over SSH
+
+This example demonstrates an in-container development loop over SSH. It deploys a small
+vLLM server with SSH access enabled, then shows how to SSH into a running replica, edit
+files in place, and restart the app to pick up the changes without redeploying or
+recreating the container.
+
+The server is launched by [`entr`](https://eradman.com/entrproject/), which watches
+the files in `/app/data` and relaunches vLLM whenever one of them changes. Every
+vLLM flag lives in [`data/vllm_args.txt`](data/vllm_args.txt), and the assistant
```
marius-baseten, line 10: For consistency also add `/app/` here or remove above (also applies to following paths).

inference-ssh/tune-vllm-args/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+It serves [Qwen/Qwen2.5-0.5B-Instruct](https://huggingface.co/Qwen/Qwen2.5-0.5B-Instruct)
+on a single L4 GPU.
+
+## 1. Set up SSH
+
+Set up SSH once per machine, using the
+[Truss CLI](https://docs.baseten.co/reference/cli/truss/overview) or the
+[Baseten CLI](https://docs.baseten.co/reference/cli/baseten/overview). Install
+whichever you prefer by following its docs, then run its setup below. For background
```
marius-baseten: "whichever you prefer" puts unneded burden on user. If I wasn't familiar with basteten stuff and see something like this I would need to ask claud "which *should* I prefer and why?" (or pre 2025 read all kind of other doc pages to figure out).
If you tell claude to follow this tutorial it will either stop autonomous work and ask the user what to do, or it will chose silently as nondeterinistic behavior - both are not great.

This is related to the other discussion about AGENTS.md vs. skills etc. I take a strong point here:

> More options is not better per se. What is better is clarity.

At least to help navigate the options. Lead with a recommendation BLUP, then in the a deferred section (expandable, fine print, "advanced subpage") document alternatives, fallbacks.

I think here we write:

> We recommend our new Basten CLI (early access), but if you want to keep using your existing Truss install or encounter issues with the CLI, you can use Truss CLI.

marius-baseten: I also think taking a clear stance on what is recommended would allow splitting this into the recommended sections first and then the alternative as an appending. Right now it's hard to read.

inference-ssh/tune-vllm-args/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+## 4. Call the model
+
+Set the model ID from the push output as an environment variable.
+
+```sh
+export MY_MODEL_ID=<model_id>
+```
+
+**Using curl**
```
marius-baseten, line 58: Need to be moved down.

inference-ssh/tune-vllm-args/data/chat_template.jinja
marius-baseten: Just read the rest, maybe that's not possible here since you use a pre-canned image/server.

marius-baseten (approved):

inference-ssh/tune-vllm-args/data/chat_template.jinja
cretz: This is more of just a contrived example showing you can adjust some args and restart vllm, the args themselves are not important. Chat template was the easiest way to show the obvious change.

inference-ssh/tune-vllm-args/README.md
cretz: There is a difference between what is watched (the container path of `/app/data`) and the file content (here in repo in `data`, linked). The `/app` prefix makes sense when talking about container paths like for watching, but it would make no sense when talking about files in this repository that we're linking.

inference-ssh/tune-vllm-args/README.md
cretz: This is temporary until `baseten` CLI is considered the stable CLI of choice. I am not sure we do recommend Baseten CLI, nor do I think this example of the SSH feature needs to be where we recommend one CLI over another. We have a whole living docs page for this at https://docs.baseten.co/reference/cli, I'm not sure we need to embed it here.

I do think once we document `baseten` CLI at https://docs.baseten.co/inference/ssh we can come back here and adjust. I don't think we need to over-focus on this one Truss example's wording on CLI choice.

inference-ssh/tune-vllm-args/README.md
cretz: Not following, can you clarify a bit? Are you saying below the API key environment variable? That is intentionally only for curl, the Baseten CLI predict approach does not need that.

marius-baseten: The Curl heading is above setting the env var, but it should be moved down to the request examaple

cretz: The env var is specific to the curl heading/section by intention. If you're not using curl, you don't need it. I updated README to clarify the API key is only needed for the curl part.

inference-ssh/tune-vllm-args/README.md
cretz: By request, `baseten` CLI has been made the only CLI in this example

inference-ssh/tune-vllm-args/README.md
cretz: By request, we have adjusted these instructions to be `baseten` CLI only and have removed all references to `truss`

